PR #1, opened 2018-03-05
Refactor declarations and remove unused requires

andreluis034:
- A couple of require/imports that weren't being used have been removed from some routes.
- All libraries loaded and their respective variable declarations set to const.
- Having a 'database_name' as the key value in the config file didn't make much sense since I believe the whole app will use the same database with different collections 
- The config file now allows you to disable/enable authentication
- In case authentication is desired you can now specify the authentication database(which can be different from admin)
- The http listen port can be specified in the config file instead of in server.js
- The source/from email address is now set to the same as the login instead of the static 'pretendonetwork@gmail.com'
- The server env(production or development) is set by npm and the server should be started with:
`npm run start` for production
`npm run start:dev` for development

### DIFF
--- a/constants.js
+++ b/constants.js
@@ -1,8 +1,3 @@
-const ENV = process.env.NODE_ENV || 'dev';
-
-let fs = require('fs'),
-    path = require('path');
-
 module.exports = {
     URL_ENDPOINTS: {
         mii: 'http://mii-secure.account.riiu.net/'
--- a/db.js
+++ b/db.js
@@ -1,13 +1,19 @@
-let mongoist = require('mongoist'),
-    config = require('./config.json'),
-    user_database_name = 'nintendo_acts',
-    user_database_collection_name = 'users',
-    user_database, user_collection;
+const mongoist = require('mongoist');
+const config = require('./config.json');
+const user_database_collection_name = 'users';
+const database = config.mongo.database
+const hostname = config.mongo.hostname
+const port     = config.mongo.port
+const auth     = config.mongo.authentication
 
-let username = config.mongo.nintendo_acts.username;
-let password = config.mongo.nintendo_acts.password;
-let hostname = config.mongo.nintendo_acts.hostname;
-let connection_string = 'mongodb://' + username + ':' + password + '@' + hostname + '/' + user_database_name + '?authSource=admin';
+let connection_string = '';
+
+if(config.mongo.use_authentication) {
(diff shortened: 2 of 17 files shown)


RedDuckss:
```
Having a 'database_name' as the key value in the config file didn't make much sense since I believe the whole app will use the same database with different collections
```
My concern with this is forcing users who may want to host a private server to use our database name. I left it as a config entry so that they may have the freedom to pick where things are stored

```
The config file now allows you to disable/enable authentication
```
Can you be specific on what this means? Do you mean database authentication, PNID authentication, header authentication, etc?

```
you can now specify the authentication database(which can be different from admin)
```
I left it as admin mostly for simplicity, since most tutorials use "admin" for the auth database (and I assumed people who have not used Mongo before would be using those tutorials)


Other than that everything looks really good! You found a lot of left over code and forgotten things that I missed,

andreluis034:
> My concern with this is forcing users who may want to host a private server to use our database name. I left it as a config entry so that they may have the freedom to pick where things are stored.
I think you've miss interpreted in the current master branch you have in db.js `user_database_name = 'nintendo_acts',`
and you do `config.mongo.nintendo_acts.hostname;`to access the config file 
do in the config file you would need to have the database name as a key value corresponding to the database name in db.js like so:
`
{
    "mongo": {
        "nintendo_acts": {
//omitted
        }
    }
}` 
now you just need to have 
`    "mongo": {
        "database": "nintendo_acts",
        //omitted
        }
}`
From what I understand the solution in the master branch allows for multiple databases to be used at once for this single application, but I don't think multiple databases are relevant in this context or am I wrong?
>Can you be specific on what this means? Do you mean database authentication, PNID authentication, header authentication, etc?

I did mean mongodb authentication yes, sorry for not specifying that

>I left it as admin mostly for simplicity, since most tutorials use "admin" for the auth database (and I assumed people who have not used Mongo before would be using those tutorials)

Sure most people, but not everyone, usually just configure a user in the admin database but I don't think it's bad to support other alternatives and have the admin database in example config should be fine

RedDuckss:
```
I think you've miss interpreted
```
You're right, I did. That's what I get for trying to read a PR as the first thing I do in the morning :P

```
I did mean mongodb authentication yes, sorry for not specifying that
```
Ahh gotcha, makes sense.

```
I don't think it's bad to support other alternatives
```
I wasn't saying it was bad, I was just explaining my initial reasoning :P I agree with this change, apologies for not making this clear

PR #7, opened 2019-06-07
Complete rewrite

RedDuckss:
Rewrote the entire server from the ground up. Not every endpoint is implemented, but the key ones required to get the console to link an account made with https://github.com/PretendoNetwork/website/tree/dev to the console as well as issue out access, refresh, and service tokens along with the address to a NEX server do work


RedDuckss:
Closed in favor of new rewrite

PR #16, opened 2021-08-20
Separate Canberra from Melbourne and Sydney

AnnoyingRain5:
Canberra does not follow daylight savings time when Melbourne and Sydney do, they should be separated to avoid issues with the time becoming desynchronized during daylight savings time.


jonbarrow:
the master branch is outdated, I started pushing updates to the dev branch a while ago and have made a lot of changes to my local account server which will be pushed there and merged into master soon. so this PR will also become outdated

PR #42, opened 2022-09-21
Added additional Config-Options for AWS.

Gandalf1783:
I have added a URL-Parameter and a Bucket-Parameter in order to make it easier to change these.

It was hard to find the url and bucket in the code, especially because the bucket is used in multiple places.

### DIFF
--- a/src/example.config.json
+++ b/src/example.config.json
@@ -18,6 +18,8 @@
     },
     "aws": {
         "spaces": {
+            "url": "nyc3.digitaloceanspaces.com",
+            "bucket": "pn-cdn",
             "key": "ACCESS_KEY",
             "secret": "ACCESS_SECRET"
         }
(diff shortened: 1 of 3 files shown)


mrjvs (approved):
lgtm, although it does have deployment breaking changes

Gandalf1783:
I can see the issues it will create.
We could overcomplicate things with default values, but I think that wont be an option.

You could technically replace it with this:
`( config.aws.spaces.bucket !== undefined ? config.aws.spaces.bucket : "pn-cdn")` (Just an example)
and test if there is an option set in the configuration, if not use the default value . That would not break anything until the values in the config.js would be overwritten.

I guess there are no direct ways to do this without breaking stuff.

jonbarrow:
Seems I closed this with a comment last night as well. This has been closed as our entire config system has changed. We have moved away from JavaScript in favor of TypeScript, moved away from json configs in favor of environment variables, s3 settings are already configurable _and_ now optional

PR #47, opened 2022-09-28
Updates and security fixes

jonbarrow:
Several security issues and general bugs fixed in this PR:

- Removed old `dockerfile` and `servers.json` files (no longer used)
- Throw 400 on XML parse error
- Remove use of `this` in database.js
- Removed all sync fs functions, replaced with async/await
- Cached files read from disk in Redis
- All database modifications now use transactions to prevent desync issues
- CDN urls now use the config
- Implement Joi for object validation
- Removed query injection when adding Discord connection
- Fixed email tokens being regenerated when password is updated
- Made NEX accounts unique based on console type _and_ PID (A WiiU NEX account and 3DS NEX account may share a PID as they are isolated accounts)
- Removed PNID/NEXAccount pre save functions in favor of more explicit function calls
- Fixed a bug where a failed token decryption would crash the server


src/server.js
```diff
@@ -70,11 +71,16 @@ app.use((error, request, response) => {
...
 
-database.connect().then(() => {
 	app.listen(port, () => {
 		logger.success(`Server started on port ${port}`);
 	});
-});
\ No newline at end of file
+}
+
+main();
```
mrjvs, line 86: I recommend doing a `.catch(err=>console.error(err))` here so you nodejs don't yell at you in newer versions.
in some versions I doesnt even log the error if you dont catch

src/services/api/routes/v1/login.js
```diff
@@ -91,12 +89,21 @@ router.post('/', async (request, response) => {
 		});
 	}
 
-	const publicKey = fs.readFileSync(`${cryptoPath}/public.pem`);
-	const hmacSecret = fs.readFileSync(`${cryptoPath}/secret.key`);
+	let publicKey= await cache.getServicePublicKey('account');
+	if (publicKey === null) {
+		publicKey = await fs.readFile(`${cryptoPath}/public.pem`);
+		await cache.setServicePublicKey('account', publicKey);
+	}
```
mrjvs, line 96: I'd recommend making the cache autoread if it doesn't exist in the cache.
that way you don't have this check everytime.
then its just `const key = await cache.getsomething('access/public.pem', 'account')`.

src/services/api/routes/v1/register.js
```diff
@@ -4,11 +4,14 @@ const fs = require('fs-extra');
...
 const hcaptcha = require('hcaptcha');
+const bcrypt = require('bcrypt');
 const { PNID } = require('../../../../models/pnid');
 const { NEXAccount } = require('../../../../models/nex-account');
 const database = require('../../../../database');
+const cache = require('../../../../cache');
 const util = require('../../../../util');
-const config = require('../../../../../config');
+const logger = require('../../../../../logger');
+const config = require('../../../../../config.json');
```
mrjvs, line 14: mayhaps setup a root path alias, you can use typescript compiler for that.
then you can just import with `import foo from '@/logger'` or `import foo from '~/logger'`. (two common alias conventions.

Your choice tho, not really neccesary, just looks way nicer and more readable if you do.
(ps: typescript compiler !== typescript language, you can use normal js files that dont use typescript language with the compiler)

src/database.js
```diff
@@ -1,18 +1,26 @@
 const mongoose = require('mongoose');
 const bcrypt = require('bcrypt');
+const joi = require('joi');
 const util = require('./util');
 const { PNID } = require('./models/pnid');
 const { Server } = require('./models/server');
-const { mongoose: mongooseConfig } = require('../config.json');
-const { uri, database, options } = mongooseConfig;
+const config = require('../config.json');
```
mrjvs, line 7: not a comment about this line, but if you're cleaning up account anyway.
want to add an example config? last time I had to get this running I had to brute force which config settings it required. (run, check error, add config, run check error add config, run, etc...)

mrjvs (changes requested):

src/services/api/routes/v1/register.js
mrjvs: actually seems like there is an easier method: https://www.npmjs.com/package/module-alias

src/database.js
jonbarrow: There should be one already? Is it outdated?

mrjvs: didnt see it, it was in the source folder. my bad

src/services/api/routes/v1/register.js
mrjvs: yes

src/server.js
jonbarrow: Fixed

mrjvs (approved):

PR #48, opened 2022-09-30
docker: Allow option to autogenerate keys

QuarkTheAwesome:
While this isn't optimal for prod, for dev environments it can be very helpful to opt-in for things to Just Work.

Off by default, can opt in via an environment variable.


jonbarrow:
Whoops I didn't see this PR until after I made changes to the entrypoint script!

QuarkTheAwesome:
Conflict is fixed!

PR #51, opened 2022-10-09
[Feature] Better config

jonbarrow:
This PR makes most config settings optional by either disabling certain features or using sensible fall-backs. Also adds the ability to load config settings from environment variables instead of a config file using a config manager now instead of using the config.json from disk everywhere


src/config-manager.js
```diff
@@ -0,0 +1,356 @@
...
+
+				set(config, keyPath, convertType ? convertType(newValue) : newValue);
+			}
+		}
+	}
+
+	// * Check for optional settings
+
+	const redisConfigValue = get(config, 'redis.client.url');
+	const redisEnvValue = get(process.env, 'PN_ACT_CONFIG_REDIS_URL');
```
QuarkTheAwesome, line 151: The flow here feels a bit weird for the `PN_ACT_PREFER_ENV_CONFIG` case, since we already read in that same variable before, thus `redisConfigValue` and `redisEnvValue` are identical.
Are we sure we want mixed env/config file options in the first place? Server usecase would be config.json, Docker usecase would be envvars...

Assuming we want it, maybe some function like..
```js
let configJson = ...;

function getValue(name) {
    if (!PN_ACT_PREFER_ENV_CONFIG && configJson...) {
        return configJson...;
    } else {
        return process.env...;
    }
}

const redis = getValue('redis.client.url');
if (!redis || redis.trim === '') {
    ...
}
set(config, ...);
```
You'd need a way to map between the dot form and envvar names, though `redis.client.url` -> `REDIS_CLIENT_URL` seems like a pretty straightforward transform.

src/cache.js
```diff
@@ -140,19 +155,38 @@ async function getServiceAESKey(name, encoding) {
...
+
+// Local CDN cache functions
+
+async function getLocalCDNFile(name, encoding) {
+	let file = await getCachedFile(`local_cdn:${name}`, encoding);
+
+	if (file === null) {
+		if (await fs.pathExists(`${LOCAL_CDN_BASE}/${name}`)) {
+			file = await fs.readFile(`${LOCAL_CDN_BASE}/${name}`, { encoding });
+			await setLocalCDNFile(name, file);
```
QuarkTheAwesome, line 181: How big are the CDN files (mii images, right?) We should make it clear that this is a dev-only feature to stop people running servers and running out of memory.

Have you checked if this is actually faster for large files? Small stuff sure, but OSes are already pretty good at disk caching these days. I'm not sure we should even bother with caching for a dev thing.

src/services/local-cdn/routes/get.js
```diff
@@ -0,0 +1,16 @@
+const router = require('express').Router();
+const cache = require('../../../cache');
+
+router.get('/*', async (request, response) => {
+	const filePath = request.params[0];
+
+	const file = await cache.getLocalCDNFile(filePath);
```
QuarkTheAwesome, line 7: :blue_heart:

src/util.js
```diff
@@ -203,20 +208,32 @@ function unpackToken(token) {
...
+			Bucket: bucket,
+			ACL: acl
+		};
+
+		await s3.putObject(awsPutParams).promise();
+	}
+}
+
+async function writeLocalCDNFile(key, data) {
+	const filePath = `${__dirname}/../cdn/${key}`;
```
QuarkTheAwesome, line 232: This should come from the config, so that I can docker-volume it out

QuarkTheAwesome (commented):
Looks sweet! Should make life much easier for local testing and maybe small deployments. I am a little unsure about the merits of mixing env and json config, but we can talk about that.

jonbarrow:
It's not so much "mixing", it's using env variables as a fall back (unless the user sets the "prefers" env variable)

This was something jvs had mentioned a while ago, allowing servers to be controlled via env variables being helpful for docker

ashquarky:
I agree that envvars are very helpful for docker, I just wonder if using a fallback on a var-by-var basis is a good idea, rather than "your whole config is a json, or your whole config is in env". Should a piecemeal config.json just be rejected?

See e.g. Mongo, which [suggests using env vars](https://github.com/docker-library/docs/blob/master/mongo/README.md#-via-docker-stack-deploy-or-docker-compose) *or* [using a config file](https://github.com/docker-library/docs/blob/master/mongo/README.md#using-a-custom-mongodb-configuration-file).

src/util.js
jonbarrow: I'm assuming you mean the path where the CDN contents is written?

src/services/local-cdn/routes/get.js
jonbarrow: :blue_heart:

src/util.js
ashquarky: Yeah, the filesystem path where they're written and read from

src/cache.js
jonbarrow: Right now the server can technically support any sized file since the `datastore` service makes use of these functions too now. But that service will be removed in the future leaving, yes, only Mii images for now. I did some light testing with different sized files and reading from memory was always faster, but I'm aware this could be a classice case of "works on my system"

As for making it clear it's a dev-only feature, when the memory cache is being used the console prints out a warning saying it's using an in-memory cache. Any ideas on how to make that more clear?

src/config-manager.js
jonbarrow: > The flow here feels a bit weird for the `PN_ACT_PREFER_ENV_CONFIG` case

I agree it's a bit redundant in this case. The later checks are done just as fail-safes for the optional settings ignoring which option the user preferred. The idea behind it is "fill out the config as best it can depending on how the user wants to load it, then try to fill in any blanks"

> Are we sure we want mixed env/config file options in the first place?

I would say yes. Giving users the option to do so feels better to me, especially considering it's entirely optional. I was also told that having configs be env variables is better for docker, which I plan to run the entire network off docker eventually too

> Assuming we want it, maybe some function like

That's not a bad setup, can definitely work something like that out

src/cache.js
ashquarky: I am specifically thinking about the case with no S3, rather than no Redis (which still uses memory proportional to the filesizes, right? or does it offload things to disk under pressure?)
> s3 file storage disabled. Using disk-based file storage. Please ensure cdn_base config or PN_ACT_CONFIG_CDN_BASE env variable is set to point to this server with the subdomain being ${config.cdn_subdomain}

is what we say right now in that case.

Maybe something like
> Both s3 and Redis are disabled. Large CDN files will use the in-memory cache, which may result in high memory use. Please enable [s3 I assume] if you're running a production server.

src/config-manager.js
ashquarky: See https://github.com/PretendoNetwork/account/pull/51#issuecomment-1272691244

jonbarrow:
Ahh I see what you mean now. Are there any downsides to doing the mixing though? Besides just some slightly more complex checks, which can be solved with some minor refactoring like you suggested? If there's no real downsides or issues to this approach then I guess my only argument is "why not"?

ashquarky:
I just think it has the potential to be confusing, though I guess the only situation where someone would use envvars at all is Docker, where you can easily control whether there's a config.json or not... So yeah, if you think it's neat enough post-refactoring, then sure.

src/cache.js
jonbarrow: Oh my apolgies I misunderstood. Redis does not dump anything to disk by default last I checked, though it can be configured to do so

That warning message is very clear, will do

jonbarrow:
The more I thought about it the more I agree that the env checks at the end were pointless. Removed them

ashquarky (approved):
Can't find the button to resolve my conversations, but all looks good to me.

PR #54, opened 2022-10-28
Added README.md for clarification

Gandalf1783:
Now on the correct branch...

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,19 @@
+# Account-Server
+
+Replacement for https://account.nintendo.net
+
+Requires a S3-compatible storage (e.g. MinIO (free)) aswell as a MongoDB server.
+
+
+Setting up:
+---
+_____
+
+1. Clone / Download this repository onto your server.
+2. Run `npm i` to install all packages.
+3. Enter the src-directory and copy the `example.config.json` and `example.servers.json` to `config.json` and `servers.json` respectively.
+4. Go ahead and edit `config.json`. Here you mainly need to enter your S3-Storage instance credentials aswell as your MongoDB connection.
+5. When done, you may try to run your instance by executing `node server.js` in the `src` folder.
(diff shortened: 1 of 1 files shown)


jonbarrow:
Thank you for the PR. However some information is wrong

1) This server is not just a replacement for `account.nintendo.net` anymore. It also replaces `nasc.nintendowifi.net`, as well as adds in our own dedicated `api` service for querying account details

2) S3 is no longer required, if an S3 configuration is not found it will default to disk-based file storage with a warning to the user

3) `servers.json` is no longer used at all, and the example file is no longer present in the `dev` branch

4) Configs can also be set using environment variables, not just the `config.json`. We support `.env` files as well for this

5) The language used seems less "formal" (for lack of a better word?) than I'd normally use. For example instead of `Go ahead and edit config.json.`, I would likely write it as something like `Open config.json and fill in your configuration settings.`

Gandalf1783:
Thanks for the update, didn't check the new changes for a while.
I will update the Readme accordingly soon.

Gandalf1783:
I have edited the Readme as far as I could according to your points.

Yet, it is not clear how to fill out the fields like the S3-Storage in the configuration if you do not wish to use it. Does it just have to be empty or the "default" values (e.g.  ACCESS_KEY and ACCESS_SECRET)? This may be intresting to add to the Readme. (Similar thing with Redis & Email)

Since I havent tested the dev-branch of this repo yet, I could not figure out what the CDN-Part of the config does. I assume this is the "API-Service for account detail queries" you talked about?

If further edits are needed, I'm happy to help.

Gandalf1783:
Don't see why this was closed without further notice?

jonbarrow:
My apologies, I thought I had closed these with comments last night. I did some cleaning up yesterday and closed some stale/obsolete PRs and issues. This was closed as the dev branch already has an updated README with setup instructions

Gandalf1783:
Thank you for clarification on both of my PRs!

PR #67, opened 2023-04-08
Update package.json

Jvr2022:
Some changes I made are: Description added Main script added start:dev script removed and replaced with a dev script that uses nodemon Keywords added License changed to MIT Homepage URL without #readme yesno and prompt devDependencies removed as they are not used ribbon script changed to `eslint

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,30 +1,35 @@
 {
   "name": "account",
   "version": "1.0.0",
-  "description": "",
-  "main": "./src/server.js",
+  "description": "Backend API for user account management",
+  "main": "src/server.js",
   "scripts": {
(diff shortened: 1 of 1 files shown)


jonbarrow:
1. We do not use MIT, our license is AGPL-3. A licnese cannot be changed like that without explict permission from every contributor, which you do not have nor are we interested in changing licenses
2. You changed the `author` property to just "Your Name Here"..?
3. You removed several key dependencies. `dicer` is part of our multipart form-data parser, `node-rsa`, and `yesno` are used to generate the crypto keys used by the game servers, and `prompt` is used in the script to create test users

Also the master branch should not me PRed into from any branch besides the dev branch. All changes should be PRed into dev before they get moved into master

PR #69, opened 2023-04-09
Update example.config.json

Jvr2022:
I improved the code by organizing the configuration parameters into different sections based on their functionality, such as HTTP, MongoDB, Redis, email, S3, hCaptcha, CDN, and website base. I also used more descriptive and readable names for the configuration parameters and nested them appropriately. This makes it easier to read and modify the code in the future.

### DIFF
--- a/example.config.json
+++ b/example.config.json
@@ -5,7 +5,8 @@
     "mongoose": {
         "connection_string": "mongodb://localhost:27017/database_name",
         "options": {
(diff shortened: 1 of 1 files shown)


jonbarrow:

1. We are moving to using TypeScript in https://github.com/PretendoNetwork/account/pull/65 which also removes the `config.json` system entirely in favor of environment variables

PR #70, opened 2023-04-09
Update create-test-user.js

Jvr2022:
I made several improvements to the code:

Added error handling for the database.connect() call using a .catch() block. This will ensure that any errors that occur during database connection are properly handled.

Refactored the callback function in prompt.get() to an async/await function for readability and ease of error handling.

Added try/catch blocks around the code that saves the new user to the database to ensure proper error handling.

Used destructuring assignment to simplify the code for retrieving values from the properties array.

Replaced the deprecated new PNID() method with the modern PNID.create() method.

### DIFF
--- a/create-test-user.js
+++ b/create-test-user.js
@@ -14,66 +14,69 @@ const properties = [
 	}
 ];
 
-prompt.get(properties, (error, { username, email, password }) => {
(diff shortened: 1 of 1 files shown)


jonbarrow:

PR #71, opened 2023-04-09
Update server.js

Jvr2022:
I made several improvements to the code:

Added error handling for database and cache connections using try-catch blocks, to ensure that the application doesn't start if these connections fail.

Changed the 404 handler to use response.sendStatus() instead of response.status().send(), which makes the code shorter and more concise.

Added a catch-all error handler at the end of the middleware chain to handle any uncaught errors in the application.

Moved the setup of middleware and routers to their own functions to make the code more modular and easier to read.

Renamed some of the variables and functions to make their purpose clearer and more descriptive.

### DIFF
--- a/src/server.js
+++ b/src/server.js
@@ -1,93 +1,76 @@
-process.title = 'Pretendo - Account';
-
-const configManager = require('./config-manager');
-
-configManager.configure();
-
 const express = require('express');
 const morgan = require('morgan');
 const xmlparser = require('./middleware/xml-parser');
 const cache = require('./cache');
 const database = require('./database');
-const util = require('./util');
 const logger = require('../logger');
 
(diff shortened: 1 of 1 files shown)


jonbarrow:

1. You have removed the config system entirely with regards to the http port, which you have previously tried to improve with a different PR, rending that useless
2. Module importing should only happen at the top of files (with regards to your `setupRouters` function)
3. Besides the error try-catch for the database connections, which are helpful, not much else is meaningful enough to warrent a PR. The changes you made seem to lean more towards a personal preference than actual improvements
4. In the 404 and 5XX handlers you completely removed the response body and headers that the console expects to be there for it's own error handling. Did you test any of this on real hardware?

PR #72, opened 2023-04-09
Update index.js

Jvr2022:
Reordered imports for better readability.
Moved the logger import to the top of the file, and added a success log message once the subdomains have been set up. Removed unnecessary comments that just repeated what the code already said. Removed the initial log messages for middleware and route setup, as they do not add value to the code. Removed the uppercase names for routes, as they do not follow JavaScript naming conventions. Simplified the registration of routes on the nnid router by chaining them. Moved the creation of the nnid router and the registration of its middleware and routes to the top of the file for better organization.

### DIFF
--- a/src/services/nnid/index.js
+++ b/src/services/nnid/index.js
@@ -1,23 +1,19 @@
-// handles "account.nintendo.net" endpoints
-
 const express = require('express');
 const subdomain = require('express-subdomain');
+const logger = require('../../../logger');
 const clientHeaderCheck = require('../../middleware/client-header');
 const cemuMiddleware = require('../../middleware/cemu');
 const pnidMiddleware = require('../../middleware/pnid');
-const logger = require('../../../logger');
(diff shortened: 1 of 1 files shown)


jonbarrow:

1. Again we are moving to TypeScript in https://github.com/PretendoNetwork/account/pull/65 which completely changes imports
2. The logs do add value to the code as if errors happen in any of those functions the logs files catch exactly where the code got to before hand, which is faster for initial analysis than reading a stack trace right away
3. You claim to have made changes to the "uppercase names for routes" but there are no changes relating to that
4. You claim the changes to the "uppercase names for routes" are for JavaScript naming conventions, yet there is no convention going against this. If anything, this is completely within spec as the route names are treated as constants which should be all uppercase and snakecase
5. You claim to have changed where certain routes are created but there are no changes relating to this. All you have done is moved where the logger is imported, removed comments, and removed logging, none of which are meaningful changes

PR #74, opened 2023-07-16
Add GetNEXData GRPC method

DaniElectra:
Returns all details of a NEX account.


src/services/grpc/account/get-nex-data.ts
```diff
@@ -0,0 +1,24 @@
...
+import { NEXAccount } from '@/models/nex-account';
+import { HydratedNEXAccountDocument } from '@/types/mongoose/nex-account';
+
+export async function getNEXData(request: GetNEXDataRequest): Promise<DeepPartial<GetNEXDataResponse>> {
+	const nexAccount: HydratedNEXAccountDocument | null = await NEXAccount.findOne({ pid: request.pid });
+
+	if (!nexAccount) {
+		throw new ServerError(
+			Status.INVALID_ARGUMENT,
+			'No NEX account found found',
```
jonbarrow, line 12: `found` is repeated here

PR #77, opened 2023-11-03
Wii U Account Information App Support

CaramelKat:
This is a draft PR for the WIP account information app on the Wii U so I can get a code review going. Currently the application is read only, and requires an WIP Inkay patch from my profile. 

Currently the following hurdles need to be overcome:
- [x] The whitelist limits the app to a single domain (the account server). This prevents the profile icon from being loaded from the CDN
- [x] Mapping timezones to the names. 3DS requests this client slide, probably easier to do this server side
- [x] Update email flow. Need to investigate how this works on 3DS.
- [x] Handling the changing of the server account level. Can we force a reconnect to the friends sever after saving and exiting the app?


src/assets/user-info-settings/index.css
```diff
@@ -0,0 +1,255 @@
...
+body {
+  font-family: Poppins, Arial, Helvetica, sans-serif;
+  font-size: 15px;
+  width: 1280px;
+  margin: 0;
+  overflow: auto;
+  box-sizing: border-box;
+  background: #eaeaea;
+}
+button:active, .server-selection input:checked + label {
```
jonbarrow, line 14: I'm only gonna comment this the one time because it happens multiple times. For consistency sake, either all blocks should have no line breaks or they should all have line breaks. Some blocks are touching and some are not

src/assets/user-info-settings/index.css
```diff
@@ -0,0 +1,255 @@
...
+}
+
+body {
+  font-family: Poppins, Arial, Helvetica, sans-serif;
+  font-size: 15px;
+  width: 1280px;
+  margin: 0;
+  overflow: auto;
+  box-sizing: border-box;
+  background: #eaeaea;
```
jonbarrow, line 12: Typical style for working with hex in this code base is all uppercase

src/assets/user-info-settings/index.css
```diff
@@ -0,0 +1,255 @@
...
+}
+button:active, .server-selection input:checked + label {
+  box-shadow: inset 0 0px 10px 0 rgba(66,45,120,.75) !important;
```
jonbarrow, line 15: For readability purposes

```suggestion
  box-shadow: inset 0 0px 10px 0 rgba(66, 45, 120, 0.75) !important;
```

These kinds of spacing issues are seen multiple times here, but also only commenting once

src/assets/user-info-settings/index.css
```diff
@@ -0,0 +1,255 @@
...
+  margin: 30px 60px 0 200px;
+}
+div.group h2 {
+  font-weight: normal;
+}
+div.group {
+  /*! background: #673DB6; */
+  /*! color: #fff; */
+  /*! padding: 4px 20px; */
+  /*! border-radius: 15px; */
```
jonbarrow, line 63: Can these be removed?

src/assets/user-info-settings/index.css
```diff
@@ -0,0 +1,255 @@
...
+  background: #791e00;
+  color: #FF3F00;
+  border-color: rgba(255, 63, 0, 0.8);
+}
+
+.account-info .tier-level-1 {
+  background: #934D4D;
+	color: #FF8484;
+	border-color: #FF8484;
+}
```
jonbarrow, line 146: Starting from here down, the spacing is all over the place

This code base uses tabs of size 4

src/assets/user-info-settings/index.js
jonbarrow, line 1: Empty file?

src/server.ts
```diff
@@ -27,7 +27,8 @@ import { config } from '@/config-manager';
 const app: express.Express = express();
 
 // START APPLICATION
-
+app.set('view engine', 'ejs');
+app.set('views', __dirname + '/views');
```
jonbarrow, line 31: Space between this and the comment below

src/services/nnid/index.ts
```diff
@@ -2,6 +2,7 @@
 
 import express from 'express';
 import subdomain from 'express-subdomain';
+import path from 'path';
```
jonbarrow, line 5: When importing modules the preferred convention is:

- Native Node modules first
- 3rd party modules
- Our modules

Also native Node modules should be imported with the `node:` prefix

src/services/nnid/routes/settings.ts
```diff
@@ -0,0 +1,42 @@
+import express from 'express';
+import { getServerByClientID, getPNIDByPID } from '@/database';
+import { HydratedServerDocument } from '@/types/mongoose/server';
+import { HydratedPNIDDocument } from '@/types/mongoose/pnid';
+import { decryptToken, unpackToken, getValueFromHeaders } from '@/util';
+import { Token } from '@/types/common/token';
+//import { GetUserDataResponse } from 'pretendo-grpc-ts/dist/account/get_user_data_rpc';
+//import { getUserAccountData } from '@/util';
```
jonbarrow, line 8: Can these be removed?

src/services/nnid/index.ts
```diff
@@ -15,25 +16,48 @@ import oauth from '@/services/nnid/routes/oauth';
...
+	return next();
+}
+
+async function setJSHeader(request: express.Request, response: express.Response, next: express.NextFunction): Promise<void> {
+	response.set('Content-Type', 'text/javascript');
+	return next();
+}
+
+async function setIMGHeader(request: express.Request, response: express.Response, next: express.NextFunction): Promise<void> {
+	response.set('Content-Type', 'application/octet-stream');
```
jonbarrow, line 37: Does this have to be this way? There's dedicated image content type headers

src/services/nnid/routes/settings.ts
```diff
@@ -0,0 +1,42 @@
...
+
+const router: express.Router = express.Router();
+
+/* GET discovery server. */
```
jonbarrow, line 12: Incorrect comment

src/services/nnid/routes/settings.ts
```diff
@@ -0,0 +1,42 @@
...
...ps,cdn.jsdelivr.net,,2|1|http,kt-imogen.b-cdn.net,,2|1|http,youtube.com,,2|http,portal.cdn.pretendo.cc,,2|https,portal.cdn.pretendo.cc,,2|http,pretendo-cdn.b-cdn.net,,2|https,pretendo-cdn.b-cdn.net,,2|https,youtube.com,,2|http,.youtube.com,,2|https,.youtube.com,,2|http,.ytimg.com,,2|https,.ytimg.com,,2|http,.googlevideo.com,,2|https,.googlevideo.com,,2|https,youtube.com,/embed/,6|https,youtube.com,/e/,6|https,youtube.com,/v/,6|https,www.youtube.com,/embed/,6|https,www.youtube.com,/e/,6|https,www.youtube.com,/v/,6|https,youtube.googleapis.com,/e/,6|https,youtube.googleapis.com,/v/,6|http,maps.googleapis.com,/maps/api/streetview,2|https,maps.googleapis.com,/maps/api/streetview,2|http,cbk0.google.com,/cbk,2|https,cbk0.google.com,/cbk,2|http,cbk1.google.com,/cbk,2|https,cbk1.google.com,/cbk,2|http,cbk2.google.com,/cbk,2|https,cbk2.google.com,/cbk,2|http,cbk3.google.com,/cbk,2|https,cbk3.google.com,/cbk,2|https,.cloudfront.net,,2|https,www.google-analytics.com,/,2|https,stats.g.doubleclick.net,,2|https,www.google.com,/ads/,2|https,ssl.google-analytics.com,,2|http,fonts.googleapis.com,,2|fonts.googleapis.com,,2|https,www.googletagmanager.com,,2|http,miiverse.cc,,2|https,miiverse.cc,,2');
```
jonbarrow, line 14: I thought you said the applet didn't respect this header? Also it's got lots of URLs that are not needed here

src/services/nnid/routes/settings.ts
```diff
@@ -0,0 +1,42 @@
...
+		response.sendStatus(504);
+		return;
+	}
+	const tokenContents: Token = unpackToken(decryptedToken);
+	const PNID: HydratedPNIDDocument | null = await getPNIDByPID(tokenContents.pid);
+
+	if(!PNID) {
+		response.sendStatus(504);
+		return;
+	}
```
jonbarrow, line 35: This code base uses spacing like this

Also not done here, but `if (!decryptedToken)` will never trigger since that function will never not return anything falsey. It does, however, throw errors. So a try-catch is needed. See `getPNIDByTokenAuth` for an example

```suggestion
	const server: HydratedServerDocument | null = await getServerByClientID('3f3928cc6f780638d360f0485cef973f', 'prod');
	const token: string | undefined = getValueFromHeaders(request.headers, 'x-nintendo-service-token');
	
	if (!server || !token) {
		response.sendStatus(504);
		return;
	}

	const aesKey: string = server?.aes_key;
	const decryptedToken: Buffer = decryptToken(Buffer.from(token, 'base64'), aesKey);

	if (!decryptedToken) {
		response.sendStatus(504);
		return;
	}

	const tokenContents: Token = unpackToken(decryptedToken);
	const PNID: HydratedPNIDDocument | null = await getPNIDByPID(tokenContents.pid);

	if (!PNID) {
		response.sendStatus(504);
		return;
	}
```

src/util.ts
```diff
@@ -108,8 +108,12 @@ export function decryptToken(token: Buffer): Buffer {
 		encryptedBody = token.subarray(4);
 	}
 
+	if(!key) {
```
jonbarrow, line 111:
```suggestion
	if (!key) {

```

jonbarrow (changes requested):

src/services/nnid/routes/people.ts
```diff
@@ -580,7 +580,8 @@ router.put('/@me', async (request: express.Request, response: express.Response):
 	const region: number = person.region ? person.region : pnid.region;
 	const countryCode: string = person.country ? person.country : pnid.country;
 	const language: string = person.language ? person.language : pnid.language;
-	const timezoneName: string = person.tz_name ? person.tz_name : pnid.timezone.name;
+	// The 3DS sends an empty object when the timezone isn't updated, so this fixes validation from failing.
+	const timezoneName: string = (person.tz_name && !!Object.keys(person.tz_name).length) ? person.tz_name : pnid.timezone.name;
```
DaniElectra, line 584: I can't seem to replicate this? But that could because my account launguage is spanish, and I remember that for some people that the NNID settings worked better on spanish?

src/services/nnid/routes/people.ts
```diff
@@ -610,9 +611,8 @@ router.put('/@me', async (request: express.Request, response: express.Response):
 	pnid.region = region;
 	pnid.timezone.name = timezoneName;
 	pnid.timezone.offset = Number(timezone.utc_offset);
-	pnid.timezone.marketing = marketingFlag;
-	pnid.timezone.off_device = offDeviceFlag;
-
+	pnid.flags.marketing = marketingFlag;
+	pnid.flags.off_device = offDeviceFlag;
```
DaniElectra, line 615: Please add a space between this and the data save

While you are here, can you also make this update some missing fields that are retreived above, like `language` and `countryCode`?

src/views/index.ejs
```diff
@@ -0,0 +1,123 @@
+<!DOCTYPE html>
+<html lang="en">
+
+<head>
+    <title>Account Settings</title>
+    <link rel="stylesheet" type="text/css" href="/v1/account-settings/css/index.css">
+    <script src="/v1/account-settings/js/index.js" "=""></script>
```
DaniElectra, line 7: There's an extra or missing quotation mark `"` here?

DaniElectra (commented):
Replying to some things about the 3DS:

> Mapping timezones to the names. 3DS requests this client slide, probably easier to do this server side

The 3DS requests the timezones from the server, from `https://account.nintendo.net/v1/api/content/time_zones/ES/es` to be specific

> Update email flow. Need to investigate how this works on 3DS

I thought this was already implemented?

src/server.ts
```diff
@@ -27,6 +27,7 @@ import { config } from '@/config-manager';
 const app: express.Express = express();
 
 // START APPLICATION
+
 app.set('view engine', 'ejs');
 app.set('views', __dirname + '/views');
 // Create router
```
jonbarrow, line 33: I had meant a newline here after `app.set('views', __dirname + '/views');` and before `app.set('views', __dirname + '/views');`

src/services/nnid/routes/people.ts
CaramelKat: I encountered this while testing on my 3DS locally. It might have been a one time bug, but it was consistently sending a empty tag when the timezone wasn't changed. I didn't check to see if this continued after restarting the app.

src/services/nnid/routes/settings.ts
```diff
@@ -44,4 +67,110 @@ router.get('/profile', async function (request: express.Request, response: expre
...
 });
 
+/**
+ * [GET]
+ * Description: Fetches the requested mii image from the CDN and send it to the client.
+ * This is required because of the strict domain whitelist in the account settings app
+ * on the Wii U.
+ */
+router.get('/mii/:pid/:face', async function (request: express.Request, response: express.Response): Promise<void> {
+	const miiImage: Buffer = await got(`https://pretendo-cdn.b-cdn.net/mii/${request.params.pid}/${request.params.face}.png`).buffer();
```
DaniElectra, line 77: Can we avoid having hardcoded domains?

CaramelKat:
This implementation is now functional as far as what is provided. I have tested everything locally, and all the values seem to be updated correctly. A lot of this code is lifted from the existing account update endpoints, since the data is all the same. While missing a few small things (mostly changing region and indicating that an email is not yet verified) I feel comfortable moving this out of a draft state if @jonbarrow doesn't find any glaring issues.

src/services/nnid/routes/settings.ts
```diff
@@ -44,4 +67,110 @@ router.get('/profile', async function (request: express.Request, response: expre
...
+	response.set('Content-Type', 'image/png');
+	response.send(miiImage);
+});
+
+/**
+ * [POST]
+ * Description: Endpoint to update the PNID from the account settings app on the Wii
+ */
+router.post('/update', async function (request: express.Request, response: express.Response): Promise<void> {
+	const server: HydratedServerDocument | null = await getServerByClientID('3f3928cc6f780638d360f0485cef973f', 'prod');
```
DaniElectra, line 87: Do we want to hardcode the server access level requested to `prod`?

src/services/nnid/routes/settings.ts
```diff
@@ -44,4 +67,110 @@ router.get('/profile', async function (request: express.Request, response: expre
...
+		if (!pnid) {
+			response.status(404);
+			response.redirect('/v1/account-settings/ui/profile');
+			return;
+		}
+
+		const gender: string = person.gender ? person.gender : pnid.gender;
+		const timezoneName: string = (person.tz_name && !!Object.keys(person.tz_name).length) ? person.tz_name : pnid.timezone.name;
+		const marketingFlag: boolean = person.email_offers ? person.email_offers : pnid.flags.marketing;
+		const offDeviceFlag: boolean = person.device_access ? person.device_access: pnid.flags.off_device;
```
DaniElectra, line 111: little nitpick, but can we make these values have the same name as on the 3DS?

src/services/nnid/routes/settings.ts
jonbarrow: `config.cdn.base_url` exists for this, to be clear

jonbarrow:
I sent you a list of all the countries and their regions as extracted from the WiiU, before we move this out of draft can support for pulling from that list be added?

example.env
```diff
@@ -19,4 +19,5 @@ PN_ACT_CONFIG_CDN_SUBDOMAIN=local-cdn
 PN_ACT_CONFIG_CDN_DISK_PATH=/home/jon/pretend-cdn
 PN_ACT_CONFIG_WEBSITE_BASE=https://example.com
 PN_ACT_CONFIG_AES_KEY=abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789
-PN_ACT_CONFIG_GRPC_API_KEY=apikey
\ No newline at end of file
+PN_ACT_CONFIG_GRPC_API_KEY=apikey
+PN_ACT_CONFIG_SERVER_ENVIRONMENT=prod
```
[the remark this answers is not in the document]
jonbarrow: This is the environment the server is assumed to be running in. It's used internally to grab the right crypto key from the database. The NNID Settings app works off a service token, like a 3rd party service does, so we need to know which key to use in order to decrypt it here. Before, it was hard-coded to `prod`, but in an effort to make this more flexible it was decided to let this be configurable in the event that the keys differ (as they should)

src/services/nnid/routes/people.ts
jonbarrow, line 584: Can you clarify this a bit please?

> The 3DS sends an empty object

You mean instead of a *string* it sends an *empty object*?

src/types/services/nnid/regions.ts
```diff
@@ -0,0 +1,38 @@
...
+	id: number;
+	name: string;
+	translations: Translations;
+	coordinates: Coordinates;
+}
+
+export interface Coordinates {
+	latitude: number;
+	longitude: number;
+}
```
jonbarrow, line 38: The `Coordinates` type doesn't seem to be used outside of the `RegionsEntity` type, so it can just be merged into it

```suggestion
	coordinates: {
		latitude: number;
		longitude: number;
	};
}
```

src/views/index.ejs
jonbarrow, line 1: [NWFX](https://github.com/PretendoNetwork/NWFX/)? 🥺

I'm kidding lol, couldn't help it

src/services/nnid/routes/people.ts
CaramelKat: Correct, since it's being sent as xml, it just sends the tag like this: `<tz_name/>` instead of the full `<tz_name>timezone/name</tz_name>`. It doesn't seem to happen consistently, but not handling it would cause the app to hang, which is why I added this additional check.

jonbarrow: Gross. Alright then lol

src/types/services/nnid/account-settings.ts
```diff
@@ -0,0 +1,9 @@
+export interface AccountSettings {
+	gender: string;
```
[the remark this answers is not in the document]
CaramelKat: Is this something we would want updated for all the gender fields in the account server? Or only for the account settings object?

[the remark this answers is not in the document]
jonbarrow: It's something that should be applied to all the gender fields (really all fields across the whole repo could use better typing), however I don't think that's in the scope of this PR. I only commented on the code you've specifically written

I plan to do a larger rework of the type system at a later date to make things a bit easier to use/read, which is when I planned on doing this, but I figured I'd mention this here for your code

jonbarrow (approved):

jonbarrow:
Looks ready to merge to me, unless anyone else has any concerns

src/services/nnid/routes/people.ts
```diff
@@ -580,6 +580,8 @@ router.put('/@me', async (request: express.Request, response: express.Response):
 	const region: number = person.region ? person.region : pnid.region;
 	const countryCode: string = person.country ? person.country : pnid.country;
 	const language: string = person.language ? person.language : pnid.language;
+	// The 3DS sends an empty object when the timezone isn't updated, so this fixes validation from failing.
+	const timezoneName: string = (person.tz_name && !!Object.keys(person.tz_name).length) ? person.tz_name : pnid.timezone.name;
```
DaniElectra, line 584: This change isn't needed anymore, as it was fixed on #84

src/services/nnid/routes/people.ts
```diff
@@ -588,7 +590,6 @@ router.put('/@me', async (request: express.Request, response: express.Response):
 	if (typeof timezoneName === 'object' && Object.keys(timezoneName).length === 0) {
 		timezoneName = pnid.timezone.name;
 	}
-
 	const marketingFlag: boolean = person.marketing_flag ? person.marketing_flag === 'Y' : pnid.flags.marketing;
```
DaniElectra, line 593: The removed line should be added back

src/services/nnid/routes/account-settings.ts
```diff
@@ -0,0 +1,212 @@
...
+
+const accountSettingsSchema = z.object({
+	gender: z.enum(['M', 'F']),
+	tz_name: z.string(),
+	region: z.coerce.number(),
+	email: z.string().email(),
+	server_selection: z.enum(['prod', 'test', 'dev']),
+	marketing_flag: z.enum(['true', 'false']).transform((value) => value === 'true'),
+	off_device_flag: z.enum(['true', 'false']).transform((value) => value === 'true'),
+});
```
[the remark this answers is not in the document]
DaniElectra: I had mentioned it since we were doing fixes in that endpoint, but we can leave it for a later PR if needed

DaniElectra (commented):
The PR probably needs to be rebased again due to the `nnid` endpoint directories being renamed to `nnas`

CaramelKat:
Closing in favor of #91

PR #78, opened 2023-11-05
Better handling of system transfered devices

DaniElectra:
If a user performs a system transfer, we can't guarantee the validity of the linked PIDs to a device, not this device's existance on the database. Add them to the database if necessary.

The NEX account check has been moved to the top so that we can verify that the PID exists on the database before adding it to a device.

### DIFF
--- a/src/middleware/nasc.ts
+++ b/src/middleware/nasc.ts
@@ -122,8 +122,8 @@ async function NASCMiddleware(request: express.Request, response: express.Respon
 			// * If a user performs a system transfer from
 			// * a console to another using a Nintendo account
 			// * during the transfer and both consoles have
-			// * a Pretendo account, the devices will be swapped
-			// * since we check it using the LFCS.
+			// * a Pretendo account, the new device won't have
+			// * the user's PID.
 			// *
 			// * So, the linked PIDs won't have the user's PID
 			// * anymore.
@@ -132,6 +132,16 @@ async function NASCMiddleware(request: express.Request, response: express.Respon
 
 				await device.save();
 			}
+		}
+
+		if (device.serial !== serialNumber) {
+			response.status(200).send(nascError('102').toString());
+			return;
+		}
+
+		if (device.mac_hash !== macAddressHash) {
+			response.status(200).send(nascError('102').toString());
+			return;
 		}
 	}
 
@@ -142,10 +152,6 @@ async function NASCMiddleware(request: express.Request, response: express.Respon
 	// *
 	// * This would make the Pretendo account to not have
 	// * a device on the database.
-	// *
-	// * TODO: With this change, now multiple devices can
-	// * have the same serial number and MAC address.
-	// * Do we want this? If not, are there other solutions?
 	if (!device && pid) {
 		device = new Device({
 			model,
--- a/src/middleware/nasc.ts
+++ b/src/middleware/nasc.ts
@@ -95,12 +95,16 @@ async function NASCMiddleware(request: express.Request, response: express.Respon
 		return;
 	}
 
-	const nexAccount: HydratedNEXAccountDocument | null = await NEXAccount.findOne({ pid });
+	let nexAccount: HydratedNEXAccountDocument | null = null;
+	if (pid) {
+		nexAccount = await NEXAccount.findOne({ pid });
 
-	if (!nexAccount || nexAccount.access_level < 0) {
-		response.status(200).send(nascError('102').toString());
-		return;
+		if (!nexAccount || nexAccount.access_level < 0) {
+			response.status(200).send(nascError('102').toString());
+			return;
+		}
 	}
+
 
 	let device: HydratedDeviceDocument | null = await Device.findOne({
 		fcdcert_hash: fcdcertHash,
@@ -164,7 +168,7 @@ async function NASCMiddleware(request: express.Request, response: express.Respon
 
 			try {
 				// Create new NEX account
-				const nexAccount: HydratedNEXAccountDocument = new NEXAccount({
+				nexAccount = new NEXAccount({
 					device_type: '3ds',
 					password
 				});
--- a/src/middleware/nasc.ts
+++ b/src/middleware/nasc.ts
@@ -124,14 +124,9 @@ async function NASCMiddleware(request: express.Request, response: express.Respon
 			// * So, the linked PIDs won't have the user's PID
 			// * anymore.
 			if (!linkedPIDs.includes(pid)) {
-				const session: mongoose.ClientSession = await databaseConnection().startSession();
-				await session.startTransaction();
-
 				device.linked_pids.push(pid);
 
-				await device.save({ session });
-
-				await session.commitTransaction();
+				await device.save();
 			}
 		}
 	}
@@ -148,9 +143,6 @@ async function NASCMiddleware(request: express.Request, response: express.Respon
 	// * have the same serial number and MAC address.
 	// * Do we want this? If not, are there other solutions?
 	if (!device && pid) {
-		const session: mongoose.ClientSession = await databaseConnection().startSession();
-		await session.startTransaction();
-
 		device = new Device({
 			model,
 			serial: serialNumber,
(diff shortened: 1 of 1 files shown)


src/middleware/nasc.ts
```diff
@@ -95,6 +95,13 @@ async function NASCMiddleware(request: express.Request, response: express.Respon
 		return;
 	}
 
+	const nexAccount: HydratedNEXAccountDocument | null = await NEXAccount.findOne({ pid });
```
DaniElectra, line 98: Actually, this may cause issues with the path for account registration, since we also have a `nexAccount` there. Any suggestions?

jonbarrow: Yes this will break registration, as it will return an error if no NEX account is found. When a 3DS connects to our servers for the first time it doesn't have a PID yet and instead sends the password it would like to use for it's NEX account. Having this here will prevent any new consoles from registering

Honestly the best solution for now might be to just stop checking if the PID is linked to the device and that's it

src/middleware/nasc.ts
```diff
@@ -108,13 +115,56 @@ async function NASCMiddleware(request: express.Request, response: express.Respon
...
-				response.status(200).send(nascError('102').toString());
-				return;
+				const session: mongoose.ClientSession = await databaseConnection().startSession();
+				await session.startTransaction();
+
+				device.linked_pids.push(pid);
+
+				await device.save({ session });
+
+				await session.commitTransaction();
```
jonbarrow, line 134: Sessions are only needed when working with multiple collections which depend on the results of previous operations. For example during registration we create both a new `NEXAccount` and `Device`. However if something fails when creating the `Device`, we do not want the `NEXAccount` to be saved, that way the client can start "fresh" again at a later time

Sessions solves this problem by allowing you to stage multiple collection operates before committing to them, and allowing you to abort all of them at once. If any operation fails or the session is aborted then no changes take place for any of the collections used in the session

src/middleware/nasc.ts
```diff
@@ -108,13 +115,56 @@ async function NASCMiddleware(request: express.Request, response: express.Respon
...
+	// * if a console that has a Pretendo account performs
+	// * a system transfer using the Nintendo account to
+	// * another that doesn't have a Pretendo account.
+	// *
+	// * This would make the Pretendo account to not have
+	// * a device on the database.
+	// *
+	// * TODO: With this change, now multiple devices can
+	// * have the same serial number and MAC address.
+	// * Do we want this? If not, are there other solutions?
```
jonbarrow, line 149: No I do not believe we want this, as it makes it much harder to differentiate between unique consoles. Does a system transfer also move the serial number?

src/middleware/nasc.ts
```diff
@@ -108,13 +115,56 @@ async function NASCMiddleware(request: express.Request, response: express.Respon
...
+			serial: serialNumber,
+			environment,
+			mac_hash: macAddressHash,
+			fcdcert_hash: fcdcertHash,
+			linked_pids: [pid]
+		});
+
+		await device.save({ session });
+
+		await session.commitTransaction();
```
jonbarrow, line 165: Same session comment as before

jonbarrow (changes requested):

jonbarrow:
It looks like @InternalLoss provided a dump on Discord showing a system transfer between 2 3DS consoles. The console makes a few requests to the NNID API during this, which might be a better place to handle this behavior

Checking this dump I answered one of my own questions that, yes, a system transfer does indeed move your serial number too

Why in fucks name Nintendo thought that was an okay move is absolutely beyon me, but whatever. Now we get to deal with this

DaniElectra:
> It looks like InternalLoss provided a dump on Discord showing a system transfer between 2 3DS consoles. The console makes a few requests to the NNID API during this, which might be a better place to handle this behavior

The issue here is that the system transfer is performed on Nintendo servers, so the console doesn't ever reach to us during the transfer. Thus, *we don't know* when a console has been transfered or not. So we have to deal with this on NASC.

src/middleware/nasc.ts
DaniElectra: I was going to suggest about searching the device using the serial number and MAC address, but since the serial number is also transfered, this puts us on a tight spot.

src/middleware/nasc.ts
DaniElectra: I can do that, once we fix the missing device edge case

jonbarrow:
That's true. This is. Not a great situation. I'm unsure how to proceed here, Nintendo seems to have made it basically impossible to determine this and I'm not exactly a fan of allowing multiple devices to have the same details

What all data, that we store, gets transferred as well? Do the certs ALSO transfer? Like the `fcdcert`?

I'm trying to find SOME WAY for us to find the "original" device here

DaniElectra:
The `fcdcert` has the LFCS of the system, which is also transfered to the other console. In fact, that's why we have this edge case of a missing device.

I believe the only constant identifier we have for a device is the MAC address, since that is tied to the hardware. I guess we could use that when searching for a device?

jonbarrow:
According to @InternalLoss on Discord the MAC is trivial to replace, and given that information I assume that MAC changes are common? I'm not sure why you'd want to be I guess it's a thing

Honestly at this point I think the best move might be to just remove the error entirely when a PID isn't linked and then move the `device.linked_pids.push(pid);` from this block

https://github.com/PretendoNetwork/account/blob/a7ac9c63cb14e0d8502954be2e9e2af59d36ebf0/src/middleware/nasc.ts#L154-L165

And instead put it where the error used to be. Basically forcing all unknown PIDs to be assumed to be "linked"

I really don't know if there's a better solution right now and I honestly have half a mind to just never allow system transfers on our servers, because this is some bullshit to deal with

DaniElectra:
Now that we know that the LFCS isn't transfered between consoles, we can probably implement this in a safe way. The current implementation is correct, but we want to figure out how to avoid duplicate serial number and MAC addresses in case a custom client interacts with the server.

jonbarrow:
> in case a custom client interacts with the server

The bigger issue imo is console bans. Since this would duplicate console details, it means there would be multiple console documents in the database to mark as banned (for NASC and NNAS). But how do we determine which ones those are? How do we know "this is *the* console" for both documents?

DaniElectra:
Since we lookup the device using the LFCS, we could trust the first interaction where we receive the LFCS and consider the serial number and MAC address we receive for the first time as "trusted". Then, on further requests, we would check if those details match, and reject the request if it doesn't

jonbarrow:
> Since we lookup the device using the LFCS, we could trust the first interaction where we receive the LFCS and consider the serial number and MAC address we receive for the first time as "trusted". Then, on further requests, we would check if those details match, and reject the request if it doesn't

Sorry, it's been a few months since I actually looked at this PR. I forgot that this PR doesn't have anything to do with linking the device registered from NASC with the details obtained from NNAS. That's where my head was at

As for using the LFCS, wouldn't those kinds of checks be unsafe? We've established that the LFCS does not move consoles, but the serial does. So then this would happen:

- 2 devices, DeviceA and DeviceB
- DeviceA has SerialCodeA linked to LFCSA on the server
- DeviceB has SerialCodeB linked to LFCSB on the server
- DeviceA is transferred to DeviceB
- DeviceB now has SerialCodeA, but LFCSB
- DeviceB goes online, and LFCSB is used to get the device details
- The server thinks it should be using SerialCodeB, but it now uses SerialCodeA
- Rejected request

Was this not the cause of this issue in the first place?

I'm gonna be honest, it might be worth decoupling the LFCS from the device entirely. But then we'd need to find a way to securely ban devices again, and NASC doesn't get sent the same certificate as NNAS

I know @zaksabeast rewrote the FRD sysmodule in Rust a while ago (https://github.com/zaksabeast/3ds-friends-sysmodule). I wonder if it might be worth trying to do the same with ACT (or whatever calls NASC) and forcing the NASC requests to always send the device certificate, just like NNAS...

jonbarrow:
Actually! @DaniElectra FRD *IS* what calls NASC! I wonder if we can do something with this??

DaniElectra:
> We've established that the LFCS does not move consoles, but the serial does

The serial number and the MAC address are NOT moved between consoles, they remain the same on each individual console

jonbarrow:
lgtm

jonbarrow (approved):

PR #80, opened 2024-01-04
services: Add CBVC endpoint

DaniElectra:
This allows any version of the 3DS internet browser to connect using Pretendo.

The `:unknown` section seems to represent an incrementing "version" number, but I wasn't able to relate this number with the version shown in the browser info.

I added this endpoint on the account server as it was very simple.


jonbarrow:
LGTM

PR #84, opened 2024-02-26
Fix 3DS profile updates causing an unhandledRejection

MatthewL246:
For example, when attempting to change the email address. The 3DS sends an empty person.tz_name, which is interpreted as an empty object. This additional error handling is necessary in this case because empty objects are truthy.

I noticed that a few people were having issues changing their email addresses in #support, and I was able to reproduce this issue. I tested this fix locally and it does work, although only checking timezoneName for emptiness is a little messy.

### DIFF
--- a/src/services/nnid/routes/people.ts
+++ b/src/services/nnid/routes/people.ts
@@ -582,7 +582,9 @@ router.put('/@me', async (request: express.Request, response: express.Response):
 	const language: string = person.language ? person.language : pnid.language;
 	let timezoneName: string = person.tz_name ? person.tz_name : pnid.timezone.name;
 
-	// Fix for 3DS sending empty person.tz_name, which is interpreted as an empty object
+	// * Fix for 3DS sending empty person.tz_name, which is interpreted as an empty object
+	// TODO - See if there's a cleaner way to do this?
+
 	if (typeof timezoneName === 'object' && Object.keys(timezoneName).length === 0) {
 		timezoneName = pnid.timezone.name;
 	}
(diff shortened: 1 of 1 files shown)


DaniElectra:
Conflicts with #77 , as it includes a similar fix

@jonbarrow Should we merge this and rebase the other PR later?

MatthewL246:
Ah, I see you've already known about this issue for a while. I didn't realize that PR was related to this. Very interesting that it works differently depending on the language!

jonbarrow:
> @jonbarrow Should we merge this and rebase the other PR later?

That's fine by me. This one seems fine as-is, and #77 is still in the review process

src/services/nnid/routes/people.ts
```diff
@@ -580,7 +580,13 @@ router.put('/@me', async (request: express.Request, response: express.Response):
 	const region: number = person.region ? person.region : pnid.region;
 	const countryCode: string = person.country ? person.country : pnid.country;
 	const language: string = person.language ? person.language : pnid.language;
-	const timezoneName: string = person.tz_name ? person.tz_name : pnid.timezone.name;
+	let timezoneName: string = person.tz_name ? person.tz_name : pnid.timezone.name;
+
+	// Fix for 3DS sending empty person.tz_name, which is interpreted as an empty object
```
jonbarrow, line 585:
```suggestion
	// * Fix for 3DS sending empty person.tz_name, which is interpreted as an empty object
	// TODO - See if there's a cleaner way to do this?

```

jonbarrow (changes requested):
Functionality-wise this is fine, just need to update the comments to match the existing ones. We use the comment syntax from [Better Comments](https://marketplace.visualstudio.com/items?itemName=aaron-bond.better-comments) now

Also adds a TODO for later

PR #87, opened 2024-04-15
Updates to NNAS and general touch ups

jonbarrow:
PR touches up some ignored parts of the codebase, including removing unnecessary type declarations and updating comments

Also includes updates to the console status middleware to link 3DS documents to their device certificates. Previously this was not possible due to NASC only getting the LFCS, and NNAS only getting the device certificate. PR #78 added a safe way to do this

This has been tested against a 3DS which did not have the device certificate linked, and was updated correctly. Existing Wii U documents are also handled correctly, however I was not able to test with a fresh Wii U as both of mine are already registered


src/cache.ts
```diff
@@ -26,12 +26,12 @@ export async function setCachedFile(fileName: string, value: Buffer): Promise<vo
 }
 
 export async function getCachedFile(fileName: string, encoding?: BufferEncoding): Promise<Buffer> {
-	let cachedFile: Buffer = Buffer.alloc(0);
+	let cachedFile = Buffer.alloc(0);
```
ItzSwirlz, line 29: ok I don't know how typescript works but can you make sure this isnt a joke or something lol

src/middleware/console-status-verification.ts
```diff
@@ -69,70 +68,90 @@ async function consoleStatusVerificationMiddleware(request: express.Request, res
...
-		if (deviceID !== certificateDeviceID) {
-			// TODO - Change this to a different error
-			response.status(400).send(xmlbuilder.create({
-				error: {
-					cause: 'Bad Request',
-					code: '1600',
-					message: 'Unable to process request'
-				}
-			}).end());
+	const certificateHash = crypto.createHash('sha256').update(request.certificate._certificate).digest('base64');
```
ItzSwirlz, line 75: Future thing to think about maybe: sha512? These are 32-bit systems (3ds and Wii U) so idk if it will impact that but 64-bit servers could probably handle it

src/cache.ts
dcbaile: That's something called type inferencing - uses the context of the assignment to determine typing.

dcbaile, line 29:
```suggestion
	let cachedFile: Buffer;
```
since cachedFile is always reassigned.

shutterbug2000 (approved):

DaniElectra (approved):
Looks good to me. According to @InternalLoss the WiFi card can be replaced on an Old 3DS, so maybe we should remove the MAC address check from NASC (we're still keeping the serial number check though)

jonbarrow:
> maybe we should remove the MAC address check from NASC

That sounds fine to me. This PR or another?

src/cache.ts
jonbarrow: `cachedFile` is not always reassigned. If Redis is being used and there is no value for the key, the `cachedFile` is not updated. Using `Buffer.alloc(0)` ensured it always had a value, of some port, in these cases

It's possible this whole pattern should be changed, but that might be better off in another PR

jonbarrow: What are you talking about?

src/middleware/console-status-verification.ts
jonbarrow: That would break compatibility with older documents. sha256 is sufficient in this case

DaniElectra:
> That sounds fine to me. This PR or another?

Considering the simplicity of the change, might aswell do it here

PR #90, opened 2024-04-19
Fix NASC datetime field

jonbarrow:
The `datetime` field in NASC responses is not the Unix timestamp. It's a date and time field in the format `YYYYMMDDHHmmSS`.

PR adds a new `nascDateTime` function to get this string and updates the `datetime` usage


DaniElectra (approved):
lgtm

PR #92, opened 2024-04-27
Update NASC errors

jonbarrow:
Removes the reuse of the ban error in the NASC middleware. Reusing it leads to some confusion when the user isn't actually banned.

Uses the custom errors 150 and 151.


DaniElectra (approved):
The code looks good, but we should document the new error codes in other places before deploying

jonbarrow:
> we should document the new error codes in other places before deploying

Any suggestions for where? I do agree with that

DaniElectra:
The errors should be added at least on Bandwidth (I'm already working on this). It would be nice to add it on the wiki too

jonbarrow:
Bandwidth is merged, this should be find now

PR #97, opened 2024-06-24
Update Docker setup

MatthewL246:
### Changes:

This updates the Dockerfile and related configuration. The DataStore upload method has been updated to use an environment variable for the secret key for consistency with the server's other configuration.

This also adds a GitHub Action for building and pushing the Docker image to Docker Hub. The image is currently available at https://hub.docker.com/r/matthewl246/pretendo-account and will be deleted when this PR is merged.

- [x] ~~What I'm implementing was an [approved issue](../issues?q=is%3Aopen+is%3Aissue+label%3Aapproved).~~ This issue was approved internally.

I have tested all of the Docker changes. I have not tested DataStore HMAC secret key usage because I am not sure how to fully test that, but I believe it should work because the key is the same format that `generate-keys.js` used.


jonbarrow:
I wouldn't worry too much about the DataStore stuff. It's legacy and not used anymore iirc. It was implemented because we didn't have an S3 source that supported the TLS version we needed at the time. But we have MinIO now

MatthewL246:
Should the datastore route and associated configuration be removed then?

jonbarrow:
That's out of scope for this PR imo

MatthewL246:
The Docker image is up! https://hub.docker.com/r/matthewl246/pretendo-account

It works fine with pretendo-docker.

.github/workflows/docker.yml
```diff
@@ -0,0 +1,45 @@
+name: Build and Publish Docker Image
+
+on:
+  push:
```
mrjvs, line 4: As wolftastic mentioned on discord before, push will probably give too many images. I think only doing master and dev for push is better.

`workflow_dispatch` will allow for the escape hatch in case a non-prod version needs to be ran on the infra.

mrjvs (changes requested):

.github/workflows/docker.yml
MatthewL246: In that case, should we only use a SHA tag for images built on branches other than master and dev? Seems like they are necessary only as a fallback if latest or edge does not apply.

mrjvs (changes requested):

.github/workflows/docker.yml
```diff
@@ -0,0 +1,47 @@
...
+
+      - name: Extract Docker metadata
+        id: meta
+        uses: docker/metadata-action@v5
+        with:
+          images: ${{ github.repository }}
+          tags: |
+            type=raw,value=latest,enable=${{ github.ref == 'refs/heads/master' }}
+            type=raw,value=edge,enable=${{ github.ref == 'refs/heads/dev' }}
+            type=sha,enable=${{ github.ref != 'refs/heads/master' && github.ref != 'refs/heads/dev' }}
```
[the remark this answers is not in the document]
MatthewL246: Ah, that makes perfect sense.

binaryoverload (approved):

mrjvs (approved):

jonbarrow:
If jvs approve, Jon hits button

PR #116, opened 2024-09-21
Remove subdomain routing for the API service

binaryoverload:
### Changes:
 - Added config to allow each route to have the domains configured
 - Added a custom middleware to do routing on full domains, not just subdomains
 - Updated the build scripts to be usable on Windows
   - Added `cross-env` which allows setting of environment variables on all platforms
   - Added `copyfiles` to copy the static assets. I've used the following command to replace all the copy scripts: `copyfiles -e \"src/**/*.ts\" -u 1 \"src/**/*\" dist`
     - This copies all files from `src/**/*` to the `dist` folder including the assets, views and timezones
     - `.ts` files are excluded
     - `-u 1` is included so the files aren't copied to `dist/src` which is the default behaviour  
 - Update the config manager to only quit after all the config has been validated
 - Upgrade S3 client to v3 to stop the warning every time


jonbarrow:
I have reservations about allowing access to the API routes from any domain. They are restricted for a reason. Would the appropriate solution not be to just specify the internal subdomain as well as the public one?

binaryoverload:
> I have reservations about allowing access to the API routes from any domain. They are restricted for a reason. Would the appropriate solution not be to just specify the internal subdomain as well as the public one?

Keep in mind that "allowed from any domain" doesn't mean that any domain will be able to use this - it will still only accessible through the correct host routing in the infra which will only allow the external domain and the internal domain

The service is not responsible for restricting the domains that can access it - that's the infra's job. In this case, the subdomain routing is only so the service can tell the difference between the different domains

jonbarrow:
> The service is not responsible for restricting the domains that can access it - that's the infra's job. In this case, the subdomain routing is only so the service can tell the difference between the different domains

It's the service's job to restrict this when the service is split into multiple sub-services like ours, though. That's why we have subdomain routing to begin with, since the account server itself houses multiple services which are only accessible at specific domains

Right now the `api` service is setup for dealing with PNID related tasks, but what if we wanted to introduce other HTTP-based internal APIs? Right now most of the internal APIs are handled by internal gRPC connections, but what if we had a situation where we needed to expose, say, the server configurations through an HTTP based API for whatever reason? Without being able to differentiate between the existing account API and this new API, since subdomain routing was removed, how would it know which service to actually use?

It also feels a bit hacky, for lack of a better word? To organize everything neatly with subdomain routing, but for the account API to just go "assume literally anything else is for the account API"?

src/config-manager.ts
```diff
@@ -76,6 +77,16 @@ export const config: Config = {
...
 	datastore: {
 		signature_secret: process.env.PN_ACT_CONFIG_DATASTORE_SIGNATURE_SECRET || ''
+	},
+	domains: {
+		api: (process.env.PN_ACT_CONFIG_DOMAINS_API || 'api.pretendo.cc').split(','),
+		assets: (process.env.PN_ACT_CONFIG_DOMAINS_ASSETS || 'assets.pretendo.cc').split(','),
+		cbvc: (process.env.PN_ACT_CONFIG_DOMAINS_CBVC || 'cbvc.cdn.pretendo.cc').split(','),
+		conntest: (process.env.PN_ACT_CONFIG_DOMAINS_CONNTEST || 'conntest.pretendo.cc').split(','),
+		datastore: (process.env.PN_ACT_CONFIG_DOMAINS_DATASTORE || 'datastore.pretendo.cc').split(','),
+		cdn: (process.env.PN_ACT_CONFIG_DOMAINS_CDN || '').split(',').filter(d => d),
```
jonbarrow, line 87: I feel like this filtering should be added to all of the domain lists tbh?

src/config-manager.ts
```diff
@@ -85,21 +96,50 @@ if (process.env.PN_ACT_CONFIG_STRIPE_SECRET_KEY) {
 	};
 }
 
+// Add the old config option for backwards compatibility
```
jonbarrow, line 99:
```suggestion
// * Add the old config option for backwards compatibility
```

We use the [Better Comments](https://marketplace.visualstudio.com/items?itemName=EdwinHuiSH.better-comments-next) comment syntax

src/config-manager.ts
```diff
@@ -1,8 +1,9 @@
 import fs from 'fs-extra';
 import mongoose from 'mongoose';
 import dotenv from 'dotenv';
-import { LOG_INFO, LOG_WARN, LOG_ERROR } from '@/logger';
-import { Config } from '@/types/common/config';
+import { LOG_INFO, LOG_WARN, LOG_ERROR, formatHostnames } from '@/logger';
+import { Config, domainServices } from '@/types/common/config';
+import isValidHostname from 'is-valid-hostname';
```
jonbarrow, line 6: We have a standard import order for imports, which goes:

- Built in modules (using the `runtime:` prefix)
- 3rd party modules
- Local modules
- Types for 3rd party modules (using `import type`)
- Types for local modules (using `import type`)
- Data file imports (such as importing JSON files)

The only real exception to this is in the router files, where after the standard module imports we add a blank newline and then import all the routes in alphabetical order

src/config-manager.ts
```diff
@@ -203,9 +243,14 @@ if (!config.stripe?.secret_key) {
...
 if (!config.datastore.signature_secret) {
 	LOG_ERROR('Datastore signature secret key is not set. Set the PN_ACT_CONFIG_DATASTORE_SIGNATURE_SECRET environment variable');
-	process.exit(0);
+	configValid = false;
 }
 if (config.datastore.signature_secret.length !== 32 || !hexadecimalStringRegex.test(config.datastore.signature_secret)) {
 	LOG_ERROR('Datastore signature secret key must be a 32-character hexadecimal string.');
-	process.exit(0);
+	configValid = false;
 }
```
jonbarrow, line 251: This isn't your fault, but since we're here. There's a missing newline between these checks, and this can (should?) be swapped into a warning instead. I can't think of any game servers which still use this, and it will be removed in the future. The deployed server doesn't even have this key set, it's hot patched to disable the `process.exit` to stop it from crashing (dirty fix I did today before leaving for an appointment)

src/middleware/host-limit.ts
```diff
@@ -0,0 +1,14 @@
+import express from 'express';
+
+export function restrictHostnames<TFn extends express.Router>(
+	allowedHostnames: string[],
+	fn: TFn
+): (req: express.Request, res: express.Response, next: () => void) => void | TFn {
+	return (req: express.Request, res: express.Response, next: () => void) => {
+		if (!allowedHostnames.includes(req.hostname)) {
+			return fn(req, res, next);
```
jonbarrow, line 9: Nit, just to keep it consistent with the rest of the codebase

```suggestion
): (request: express.Request, response: express.Response, next: () => void) => void | TFn {
	return (request: express.Request, response: express.Response, next: () => void) => {
		if (!allowedHostnames.includes(request.hostname)) {
			return fn(request, response, next);

```

Also is there a reason to use this rather than something like https://www.npmjs.com/package/vhost, which Express already provides? Though vhost hasn't had any updates in years, and seems to have [issues with reverse proxies](https://github.com/expressjs/vhost/issues/20), so maybe it's not the best idea

jonbarrow (commented):
Looks great, thank you! Just a few nits tbh

src/config-manager.ts
binaryoverload: Yup, I'll add to the main processing block

src/config-manager.ts
binaryoverload: Noted, I will correct this

You will probably want to look into the eslint rule [import/order](https://github.com/import-js/eslint-plugin-import/blob/main/docs%2Frules%2Forder.md) to allow people to lint and fix this as they go along

src/middleware/host-limit.ts
binaryoverload: It looks like vhost doesn't support multiple domains outside of constructing a regex that matches all the domains

In situations like this where the logic is simple, I would rather implement it myself and have full control - especially due to the sort of stuff you mentioned with outdated code and issues

src/config-manager.ts
jonbarrow: I had no idea that was a thing, will definitely look into it tomorrow

src/middleware/host-limit.ts
jonbarrow: Very fair 👍

jonbarrow (approved):
LGTM

PR #128, opened 2024-12-12
Return the PNID's Stripe tier name and level from the PNID details API

MatthewL246:
This is a necessary for PretendoNetwork/website#345

### Changes:

This modifies the `GET https://api.pretendo.cc/v1/user` API to return the PNID's Stripe tier name and level. This is necessary for Discourse group syncing because the Discourse SSO route on the website calls this API to retreive account details, and the tier name/level will be needed to sync the supporter groups.

### DIFF
--- a/src/services/api/routes/v1/user.ts
+++ b/src/services/api/routes/v1/user.ts
@@ -62,6 +62,10 @@ router.get('/', async (request: express.Request, response: express.Response): Pr
 		connections: {
 			discord: {
 				id: pnid.connections.discord.id
+			},
+			stripe: {
+				tier_name: pnid.connections.stripe.tier_name,
+				tier_level: pnid.connections.stripe.tier_level
 			}
 		}
 	});


src/services/api/routes/v1/user.ts
```diff
@@ -62,6 +62,10 @@ router.get('/', async (request: express.Request, response: express.Response): Pr
 		connections: {
 			discord: {
 				id: pnid.connections.discord.id
+			},
+			stripe: {
+				tier_name: pnid.connections.stripe.tier_name,
+				tier_level: pnid.connections.stripe.tier_level
```
binaryoverload, line 68: I don't know what the types are here, but I reckon `connections`, and `stripe` can be undefined, which would cause an error.

I would suggest either:
a) use a ternary to say `stripe: pnid.connections?.stripe ? { ... } : null`
b) use optional chaining on the properties, i.e `pnid.connections?.stripe?.tier_name`

Option a) would give the `stripe` object the possibility to be null, and option b) would ensure the object would always be present but with possibly null values.

binaryoverload (changes requested):

PR #143, opened 2025-01-18
Mii uploads: use config for bucket name

### LINKED ISSUE #137: [Bug]: Dead stuck on attempt to download Mii avatar from pretendo-cdn.b-cdn.net via TLS1.1
### Checked Existing

### What happened?

**I am not sure on which repo is suitable this issue to be posted, but I'll post it here, as it affects many stuff.**

I found this bug three days ago, and I investigated it deeply (all my thinks and proofs I posted at Discord's support topic, but I want to get a proper feedback about this from more technically-experienced folks). First time I met that attempt to enter NNID settings or open Miiverse leads a dead stuck on attempt to TLS 1.1 handshake with the `pretendo-cdn.b-cdn.net` host. This happens from NNID and from Miiverse on attempt to download a Mii avatar (`standard.tga`).

The host itself works just fine at:
- 3DS browser when opened directly (both plain HTTP and HTTPS).
- Works just fine when pulling by CURL with any TLS version requested from command line (TLS 1.0, TLS 1.1, TLS 1.2).
- It works just fine when opening in any regular browsers on PCs and mobiles.
- It works with plain HTTP (Miiverse used it via plain HTTP only as I saw in the traffic capture).

It fails at:
- NNID settings (leads a dead stuck)
- Miiverse on load (leads a dead stuck)
- Nginx of version that available on CentOS 7 when using HTTPS host (not plain HTTP) as a reverse proxy destination (leads a dead stuck).

Also, I do see that you are migrating everything to a new CDN, and I see the Miiverse no longer takes resources from the problematic host (`pretendo-cdn.b-cdn.net`), and they loads just fine. The last thing is left the Mii avatars source that still refer the problematic host.

Mii avatar gets downloaded by Miiverse or NNID settings on load successfully, or it should report an error that server is inacessible.

### Steps to reproduce?

1. Connect 3DS to any network.
2. Try to login existing account via NNID settings (it's a chance you will get a stuck quickly here).
3. Additionally, try to open Miiverse and try to login. It also stucks here too.

According to capture traffic, it stops any traffic exchanges quickly after 3DS sent the ClientHello:
Image

## And rigth now there are three known workarounds how to escape this dead stuck:
- **Easy:** When 3DS gets stuck, do cause an intentional disconnection of 3DS from access point by one of ways: on o3DS, toggle the Wireless switch off and on, or replug router/access-point, or via router/access-point's web setup block the access, and unlock it back.

- **Average:** At the router/access-point, set up a static DNS record that will set `pretendo-cdn.b-cdn.net` host to `127.0.0.1`, or to any invalid IP address that should quickly fail. This allows to escape the stuck, but **it leads a background effect** of inability to refresh the Mii avatar.

- **Expert:** Set up a small home server (or just use your PC), and install the Nginx web server. It should force any traffic that goes to `pretendo-cdn.b-cdn.net` to be converted into plain HTTP and avoid the TLS1.1 handshake of death. You need to configure it to use Reverse Proxy by making virtual host with server_name equal to `pretendo-cdn.b-cdn.net mii.olv.pretendo.cc` with next setup:
```
location / {
    proxy_set_header  Host $host;
    proxy_pass        http://93.189.63.148/;
    proxy_redirect    off;
    proxy_buffering off;
    client_max_body_size 512m;
    proxy_connect_timeout 90;
    proxy_send_timeout 300;
    proxy_read_timeout 300;
    tcp_nodelay on;

    allow all;
}
```
And, make both HTTP and HTTPS (it's fine with self-signed or expired certificate! Also! Don't enable HTTP2, it must be HTTP1.1, otherwise, 3DS won't connect it, and this will work same as method above, but we want make it work, right?). It's important that it should run server on 80 and 443 ports. HTTPS setup should use next settings:
```
ssl_session_cache shared:SSL:10m;
ssl_session_timeout  5m;
ssl_protocols TLSv1 TLSv1.1 TLSv1.2;
ssl_ciphers @SECLEVEL=0:ALL;
ssl_prefer_server_ciphers   on;
```
Otherwise, it will fail, especially if OpenSSL 3.x is used. With OpenSSL 1.x it's fine to not add the `@SECLEVEL=0:` line. And as final step, add at your router the static DNS record that points to your home server/PC that runs Nginx: it should contain the domain `pretendo-cdn.b-cdn.net` and IP address of your home server / PC that runs Nginx. Before to test on 3DS, ensure it works with Curl and/or browser locally by the given domain. It's required to flush DNS cache to ensure valid address will be used.

InternalLoss:
Partially resolves #137

### Changes:

This PR updates the static `pn-cdn` bucket name used to store Mii images to one defined by config - once merged, this allows us to migrate to Cloudflare R2 cleanly. This also means that users who run their own instance of account will not require a bucket explicitly named `pn-cdn`.

To enable this, I have updated the environment config file, as well as a check to ensure a bucket name is present, and updated the bucket name in src/models/pnid.ts

NB: I have not been able to test these changes locally at this time, but have used existing logic to base my addit


binaryoverload (approved):
LGTM

PR #111, opened 2024-08-09
Update some authentication related checks

jonbarrow:
Resolves #XXX

### Changes:

These changes have been hot-patched in the prod servers for a long time now. This PR simply brings them upstream. Changes include:

- Only allowing `Basic` auth in some endpoints
- Validating that an access token came from a console by checking it's system type
- Validating the device certificate padding section

Marking as draft for now since I hate the way the "check the system type" feature was implemented. It's a hack, and I want to change it. There's also a couple open issues now directly relating to tokens which we may want to implement here first before merging? Unsure, would like some opinions.

- https://github.com/PretendoNetwork/account/issues/107
- https://github.com/PretendoNetwork/account/issues/108

- [x] I have read and agreed to the [Code of Conduct](https://github.com/PretendoNetwork/Pretendo/blob/master/.github/CODE_OF_CONDUCT.md).
- [x] I have read and complied with the [contributing guidelines](https://github.com/PretendoNetwork/Pretendo/blob/master/.github/CONTRIBUTING.md).
- [ ] What I'm implementing was an [approved issue](../issues?q=is%3Aopen+is%3Aissue+label%3Aapproved).
- [x] I have tested all of my changes.

PR #152, opened 2025-01-29
Fix for Invalid Service Token error - Refresh Token Duration same as Access Token

binaryoverload:
Resolves 115-5016 error occurring in Juxt 

### Changes:
- Centralises the OAuth token generation to reduce code duplication 
- Increase refresh token duration to 30 days, so when an access token expires the refresh token can be used to auth
- Moves token and system type to an "enum" (not an actual enum, but string based value) for clarity and safety

This issue was raised by imora7024 in the Discord. The problem was getting the error 115-5016 (Invalid Service Token) which, through a Charles dump, was correlated to an OAuth request failing with token auth, and then trying again with password auth.

### DIFF
--- a/src/database.ts
+++ b/src/database.ts
@@ -112,7 +112,7 @@ export async function getPNIDByNNASAccessToken(token: string): Promise<HydratedP
 		const unpackedToken = unpackToken(decryptedToken);
 
 		// * Return if the system type isn't Wii U (NNAS) and the token type isn't "OAuth Access"
-		if (unpackedToken.system_type !== 1 || unpackedToken.token_type !== 1) {
+		if (unpackedToken.system_type !== 'WIIU' || unpackedToken.token_type !== 'OAUTH_ACCESS') {
 			return null;
 		}
 
@@ -142,7 +142,7 @@ export async function getPNIDByNNASRefreshToken(token: string): Promise<Hydrated
 		const unpackedToken = unpackToken(decryptedToken);
 
 		// * Return if the system type isn't Wii U (NNAS) and the token type isn't "OAuth Refresh"
-		if (unpackedToken.system_type !== 1 || unpackedToken.token_type !== 2) {
+		if (unpackedToken.system_type !== 'WIIU' || unpackedToken.token_type !== 'OAUTH_REFRESH') {
 			return null;
 		}
 
@@ -172,7 +172,7 @@ export async function getPNIDByAPIAccessToken(token: string): Promise<HydratedPN
 		const unpackedToken = unpackToken(decryptedToken);
 
 		// * Return if the system type isn't API (REST and gRPC) and the token type isn't "OAuth Access"
-		if (unpackedToken.system_type !== 3 || unpackedToken.token_type !== 1) {
+		if (unpackedToken.system_type !== 'API' || unpackedToken.token_type !== 'OAUTH_ACCESS') {
 			return null;
 		}
 
@@ -202,7 +202,7 @@ export async function getPNIDByAPIRefreshToken(token: string): Promise<HydratedP
 		const unpackedToken = unpackToken(decryptedToken);
 
 		// * Return if the system type isn't API (REST and gRPC) and the token type isn't "OAuth Refresh"
-		if (unpackedToken.system_type !== 3 || unpackedToken.token_type !== 2) {
+		if (unpackedToken.system_type !== 'API' || unpackedToken.token_type !== 'OAUTH_REFRESH') {
 			return null;
 		}
 
--- a/src/database.ts
+++ b/src/database.ts
@@ -13,6 +13,7 @@ import { PNIDProfile } from '@/types/services/nnas/pnid-profile';
 import { ConnectionData } from '@/types/services/api/connection-data';
 import { ConnectionResponse } from '@/types/services/api/connection-response';
 import { DiscordConnectionData } from '@/types/services/api/discord-connection-data';
+import { SystemType, TokenType } from '@/types/common/token';
 
 const connection_string = config.mongoose.connection_string;
 const options = config.mongoose.options;
@@ -112,7 +113,7 @@ export async function getPNIDByNNASAccessToken(token: string): Promise<HydratedP
 		const unpackedToken = unpackToken(decryptedToken);
 
 		// * Return if the system type isn't Wii U (NNAS) and the token type isn't "OAuth Access"
-		if (unpackedToken.system_type !== 'WIIU' || unpackedToken.token_type !== 'OAUTH_ACCESS') {
+		if (unpackedToken.system_type !== SystemType.WIIU || unpackedToken.token_type !== TokenType.OAUTH_ACCESS) {
 			return null;
 		}
 
@@ -142,7 +143,7 @@ export async function getPNIDByNNASRefreshToken(token: string): Promise<Hydrated
 		const unpackedToken = unpackToken(decryptedToken);
 
 		// * Return if the system type isn't Wii U (NNAS) and the token type isn't "OAuth Refresh"
-		if (unpackedToken.system_type !== 'WIIU' || unpackedToken.token_type !== 'OAUTH_REFRESH') {
+		if (unpackedToken.system_type !== SystemType.WIIU || unpackedToken.token_type !== TokenType.OAUTH_ACCESS) {
 			return null;
 		}
 
@@ -172,7 +173,7 @@ export async function getPNIDByAPIAccessToken(token: string): Promise<HydratedPN
 		const unpackedToken = unpackToken(decryptedToken);
 
 		// * Return if the system type isn't API (REST and gRPC) and the token type isn't "OAuth Access"
-		if (unpackedToken.system_type !== 'API' || unpackedToken.token_type !== 'OAUTH_ACCESS') {
+		if (unpackedToken.system_type !== SystemType.API || unpackedToken.token_type !== TokenType.OAUTH_ACCESS) {
 			return null;
 		}
 
@@ -202,7 +203,7 @@ export async function getPNIDByAPIRefreshToken(token: string): Promise<HydratedP
 		const unpackedToken = unpackToken(decryptedToken);
 
 		// * Return if the system type isn't API (REST and gRPC) and the token type isn't "OAuth Refresh"
-		if (unpackedToken.system_type !== 'API' || unpackedToken.token_type !== 'OAUTH_REFRESH') {
+		if (unpackedToken.system_type !== SystemType.API || unpackedToken.token_type !== TokenType.OAUTH_REFRESH) {
 			return null;
 		}
 
--- a/src/models/server.ts
+++ b/src/models/server.ts
@@ -1,7 +1,6 @@
 import { Schema, model } from 'mongoose';
 import uniqueValidator from 'mongoose-unique-validator';
 import { IServer, IServerMethods, ServerModel } from '@/types/mongoose/server';
-import type { SystemType } from '@/types/common/token';
 
 const ServerSchema = new Schema<IServer, ServerModel, IServerMethods>({
 	client_id: String,
@@ -20,8 +19,3 @@ const ServerSchema = new Schema<IServer, ServerModel, IServerMethods>({
 ServerSchema.plugin(uniqueValidator, { message: '{PATH} already in use.' });
 
 export const Server = model<IServer, ServerModel>('Server', ServerSchema);
-
-export const serverDeviceToSystemType: Record<number, SystemType> = {
-	1: 'WIIU',
-	2: '3DS'
-};
--- a/src/models/server.ts
+++ b/src/models/server.ts
@@ -1,6 +1,7 @@
 import { Schema, model } from 'mongoose';
 import uniqueValidator from 'mongoose-unique-validator';
 import { IServer, IServerMethods, ServerModel } from '@/types/mongoose/server';
+import type { SystemType } from '@/types/common/token';
 
 const ServerSchema = new Schema<IServer, ServerModel, IServerMethods>({
 	client_id: String,
@@ -19,3 +20,8 @@ const ServerSchema = new Schema<IServer, ServerModel, IServerMethods>({
 ServerSchema.plugin(uniqueValidator, { message: '{PATH} already in use.' });
 
 export const Server = model<IServer, ServerModel>('Server', ServerSchema);
+
+export const serverDeviceToSystemType: Record<number, SystemType> = {
+	1: 'WIIU',
+	2: '3DS'
+};
--- a/src/services/grpc/api/login.ts
+++ b/src/services/grpc/api/login.ts
@@ -4,6 +4,7 @@ import bcrypt from 'bcrypt';
 import { getPNIDByUsername, getPNIDByAPIRefreshToken } from '@/database';
 import { nintendoPasswordHash, generateOAuthTokens} from '@/util';
 import type { HydratedPNIDDocument } from '@/types/mongoose/pnid';
+import { SystemType } from '@/types/common/token';
 
 export async function login(request: LoginRequest): Promise<DeepPartial<LoginResponse>> {
 	const grantType = request.grantType?.trim();
@@ -43,7 +44,7 @@ export async function login(request: LoginRequest): Promise<DeepPartial<LoginRes
 	}
 
 	try {
-		const tokenGeneration = generateOAuthTokens('API', pnid, { refreshExpiresIn: 14 * 24 * 60 * 60 }); // * 14 days
+		const tokenGeneration = generateOAuthTokens(SystemType.API, pnid, { refreshExpiresIn: 14 * 24 * 60 * 60 }); // * 14 days
 
 		return {
 			accessToken: tokenGeneration.accessToken,
--- a/src/services/grpc/api/login.ts
+++ b/src/services/grpc/api/login.ts
@@ -55,13 +55,13 @@ export async function login(request: LoginRequest): Promise<DeepPartial<LoginRes
 
 	try {
 		const systemType = 0x3; // * API
-		const { accessToken, refreshToken, accessTokenExpiresInSecs } = generateOAuthTokens(systemType, pnid);
+		const tokenGeneration = generateOAuthTokens(systemType, pnid, { refreshExpiresIn: 14 * 24 * 60 * 60 }); // * 14 days
 
 		return {
-			accessToken: accessToken,
+			accessToken: tokenGeneration.accessToken,
 			tokenType: 'Bearer',
-			expiresIn: accessTokenExpiresInSecs,
-			refreshToken: refreshToken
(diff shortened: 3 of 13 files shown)


jonbarrow:
This should probably target https://github.com/PretendoNetwork/account/pull/111 rather than dev, which already aims to update several authentication systems and already has reworks for the token generation

DaniElectra:
I feel 30 days is a bit too much? The console will ask for a new token on a fresh boot anyway. I think I would put it between 1 and 7 days

jonbarrow:
The official servers likely only do 24 hours in NNAS. I just replayed an oauth request I had in Charles from 4 days ago but swapped from password auth to refresh token auth, and it said the token was invalid already. It's definitely not 30 days, but doesn't seem to be 7 either

For our own tokens for the website and stuff we can raise that, but for NNAS we should keep it at 24 like the official servers

binaryoverload:
> This should probably target #111 rather than dev, which already aims to update several authentication systems and already has reworks for the token generation

From what I can see, the PR doesn't change any token generation code?

binaryoverload:
> The official servers likely only do 24 hours in NNAS. I just replayed an oauth request I had in Charles from 4 days ago but swapped from password auth to refresh token auth, and it said the token was invalid already. It's definitely not 30 days, but doesn't seem to be 7 either
> 
> For our own tokens for the website and stuff we can raise that, but for NNAS we should keep it at 24 like the official servers

Happy to update to use 24 hours for NNAS 👍

Happy for 30 days for website tokens?

jonbarrow:
> From what I can see, the PR doesn't change any token generation code?

You're right, generation was the wrong word to use there. The PR does not change the structure of tokens, but it changes how routes utilize tokens since now there's explicit handling of each token type in `database.ts` (to prevent things like website tokens from being used in NNAS, and such)

The main point being, #111 is already designed for updating auth/security and these are also auth changes, so these changes should probably target the existing PR to keep everything in sync and in one place. There's already a PR for these kinds of changes, so it makes more sense imo to keep working off that one

The existing PR is nearly ready to be merged, the last thing I wanted to do was rework the middleware some to add more authentication checks. I'm still bouncing the ideas in my head right now for how to go about it (since the Wii U only sends certain data like the device certificate in some routes, but not others whereas the 3DS sends it every time, so I have to see what can be checked where and when to bypass certain checks, etc. etc.), but I should be able to finish that out by friday

> Happy for 30 days for website tokens?

I don’t have a strong opinion either way for this. Dani mentioned thinking this sounds too long, which I do agree 30 days IS a *bit* long, but again no strong opinion. What is the standard for this?

binaryoverload:
Sure, I'll rebase off your branch then if it's going to be sorted this week 👍

I just wanted to avoiding PRing into a heavily in-progress PR, which I thought it was due to the TODOs. Since that's not the case, I'll get it rebased and re-targetted


> What is the standard for this?
In my experience 30 days is the usual for refresh token duration

There is an argument that we could set it to be something like 14 days, since the refresh token gets reset on every auth

Since you feel 30 days is a bit long, I say we reduce it to 14 days which I'd say is plenty

DaniElectra:
My concern of 30 days was only on NNAS, I'm fine with doing 30 days everywhere else

src/models/server.ts
```diff
@@ -18,4 +19,9 @@ const ServerSchema = new Schema<IServer, ServerModel, IServerMethods>({
...
 ServerSchema.plugin(uniqueValidator, { message: '{PATH} already in use.' });
 
-export const Server = model<IServer, ServerModel>('Server', ServerSchema);
\ No newline at end of file
+export const Server = model<IServer, ServerModel>('Server', ServerSchema);
+
+export const serverDeviceToSystemType: Record<number, SystemType> = {
+	1: 'WIIU',
+	2: '3DS'
+};
```
jonbarrow, line 27: The `models` folder is really only supposed to be used for Mongoose-specific code. I'm not sure this is the best place for this (for example if the use of strings was dropped in favor of number enums like it was before, then that would more nicely fit in the `types` folder)

src/services/grpc/api/login.ts
```diff
@@ -16,74 +15,43 @@ export async function login(request: LoginRequest): Promise<DeepPartial<LoginRes
...
 
 	if (grantType === 'password') {
-		pnid = await getPNIDByUsername(username!); // * We know username will never be null here
+		if (!username) throw new ServerError(Status.INVALID_ARGUMENT, 'Invalid or missing username');
+		if (!password) throw new ServerError(Status.INVALID_ARGUMENT, 'Invalid or missing password');
 
-		if (!pnid) {
-			throw new ServerError(Status.INVALID_ARGUMENT, 'User not found');
-		}
+		pnid = await getPNIDByUsername(username); // * We know username will never be null here
```
jonbarrow, line 24: This comment is useless here, it only existed because we were using `!` and TypeScript was throwing a fit

src/services/grpc/api/login.ts
```diff
@@ -16,74 +15,43 @@ export async function login(request: LoginRequest): Promise<DeepPartial<LoginRes
...
-
-	if (grantType === 'refresh_token' && !refreshToken) {
-		throw new ServerError(Status.INVALID_ARGUMENT, 'Invalid or missing refresh token');
-	}
-
 	let pnid: HydratedPNIDDocument | null;
 
 	if (grantType === 'password') {
-		pnid = await getPNIDByUsername(username!); // * We know username will never be null here
+		if (!username) throw new ServerError(Status.INVALID_ARGUMENT, 'Invalid or missing username');
```
jonbarrow, line 21: Style nit for the common config, we always use full blocks

src/services/grpc/api/login.ts
```diff
@@ -16,74 +15,43 @@ export async function login(request: LoginRequest): Promise<DeepPartial<LoginRes
...
 
-		if (!pnid) {
-			throw new ServerError(Status.INVALID_ARGUMENT, 'User not found');
-		}
+		pnid = await getPNIDByUsername(username); // * We know username will never be null here
+
+		if (!pnid) throw new ServerError(Status.INVALID_ARGUMENT, 'User not found');
 
-		const hashedPassword = nintendoPasswordHash(password!, pnid.pid); // * We know password will never be null here
+		const hashedPassword = nintendoPasswordHash(password, pnid.pid); // * We know password will never be null here
```
jonbarrow, line 28: Same comment as earlier about the comment being useless here

src/types/common/token.ts
```diff
@@ -1,8 +1,59 @@
...
+	'WIIU': 1,
+	'3DS': 2,
+	'API': 3
+} as const;
+export type SystemType = keyof typeof SystemTypes;
+
+export function getSystemTypeFromValue(type: number): SystemType | undefined {
+	const keys = Object.keys(SystemTypes) as SystemType[];
+	return keys.find((key) => SystemTypes[key] === type);
+}
```
jonbarrow, line 25: The `types` folder is really only supposed to be for type defs, I'm not sure we should be putting helper functions in here?

src/util.ts
```diff
@@ -10,9 +10,8 @@ import crc32 from 'buffer-crc32';
 import crc from 'crc';
 import { sendMail } from '@/mailer';
 import { config, disabledFeatures } from '@/config-manager';
-import { TokenOptions } from '@/types/common/token-options';
-import { Token } from '@/types/common/token';
-import { IPNID, IPNIDMethods } from '@/types/mongoose/pnid';
+import { getSystemTypeFromValue, getTokenTypeFromValue, OAuthTokenGenerationResponse, OAuthTokenOptions, SystemType, SystemTypes, Token, TokenOptions, TokenTypes } from '@/types/common/token';
+import { HydratedPNIDDocument, IPNID, IPNIDMethods } from '@/types/mongoose/pnid';
```
jonbarrow, line 14: These aren't using `import type`, I thought the eslint stuff added that? Or am I mistaken?

src/util.ts
```diff
@@ -53,15 +52,55 @@ export function nintendoBase64Encode(decoded: string | Buffer): string {
...
+		access_level: pnid.access_level,
+		expire_time: BigInt(Date.now() + (accessTokenExpiresInSecs * 1000)) // * 1 hour
+	};
+
+	const refreshTokenOptions: TokenOptions = {
+		system_type: systemType,
+		token_type: 'OAUTH_REFRESH',
+		pid: pnid.pid,
+		access_level: pnid.access_level,
+		expire_time: BigInt(Date.now() + (refreshTokenExpiresInSecs * 1000)) // * 30 days
```
jonbarrow, line 72: The comments here are kinda misleading since `options` can change this, and also the refresh token time is 1 day by default not 30 anyway?

jonbarrow (changes requested):

src/services/nnas/routes/provider.ts
```diff
@@ -89,16 +91,28 @@ router.get('/service_token/@me', async (request: express.Request, response: expr
 		return;
 	}
 
-	const tokenOptions = {
-		system_type: server.device,
-		token_type: 0x4, // * Service token
+	const systemType = serverDeviceToSystemType[server.device];
+	if (!systemType) {
```
jonbarrow, line 95: Forgot this in my last review, but another style nit for the common eslint stuff. We typically have white space separating variable declarations and blocks of code

src/util.ts
binaryoverload: Eslint is an open PR, when this all gets merged eslint will tidy it up

src/types/common/token.ts
binaryoverload: It's something I do to keep them local to the things they are relevant to - since they aren't generic helper methods

src/models/server.ts
binaryoverload: I'll move it to types 👍

src/services/nnas/routes/provider.ts
```diff
@@ -89,16 +91,28 @@ router.get('/service_token/@me', async (request: express.Request, response: expr
...
+		response.send(xmlbuilder.create({
+			errors: {
+				error: {
+					code: '1021',
+					message: 'The requested game server was not found'
+				}
+			}
+		}).end());
+	}
```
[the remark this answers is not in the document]
binaryoverload: This is specially for conversion between the number system type and the new string enum

It felt like the best error to throw, but happy to change it to something else

src/database.ts
```diff
@@ -112,7 +112,7 @@ export async function getPNIDByNNASAccessToken(token: string): Promise<HydratedP
 		const unpackedToken = unpackToken(decryptedToken);
 
 		// * Return if the system type isn't Wii U (NNAS) and the token type isn't "OAuth Access"
-		if (unpackedToken.system_type !== 1 || unpackedToken.token_type !== 1) {
+		if (unpackedToken.system_type !== 'WIIU' || unpackedToken.token_type !== 'OAUTH_ACCESS') {
```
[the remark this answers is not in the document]
binaryoverload: Magic numbers bad, using string enums is far more descriptive and allows it to also be typesafe

src/services/grpc/api/login.ts
binaryoverload: Will be linted later

src/services/nnas/routes/provider.ts
binaryoverload: Eslint hasn't been merged yet

I also personally disagree, these lines of code are directly related so I feel they should be grouped

src/services/grpc/api/login.ts
jonbarrow: I'm not sure what this response means. I made the comment because I was under the impression that this was not part of the common config (hence why I said "for the common config"). It's not clear to me what the resolution here is based on this response. Is it missing and will be added later? Does it exist and the linter will just be ran later? Something else?

binaryoverload: We agreed previously that if it's not formatted by the linter, review comments won't be made on it.

The eslint PR #151 hasn't been merged yet, so this will be formatted in accordance with the linter

binaryoverload: This has been resolved in https://github.com/PretendoNetwork/common-configs/pull/2

binaryoverload: *More like https://github.com/PretendoNetwork/common-configs/commit/3399c72dfa21995afc7b0c59a9f790ac05196ea9*

jonbarrow: > We agreed previously that if it's not formatted by the linter, review comments won't be made on it

I know you said you misunderstood on Discord, but just to clarify (and to make the clarification public): comments like this are mean't only to bring up styling examples to be added to the common config, not necessarily to fix them in the current PR (though that is also welcomed). I figured I would save making an issue and just bring it up here with a real code example. Would you prefer issues on the common config repo directly instead?

src/util.ts
jonbarrow: Perfect, I thought that was already brought in so I was just confused in that regard :+1: I thought maybe eslint just wasn't working for some reason

src/database.ts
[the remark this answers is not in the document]
jonbarrow: > Typescript enums shouldn't be used for many reasons

I keep seeing this sentiment from people in the TypeScript community but I've never seen anyone actually explain it outside of "enum bad" or something silly like "they allow mixed type values" which feels more like a failing of the implementation/developer for using them like that and not a failing of the feature (like just don't do that?), or the "issues" are things that just don't apply to us such as "they increase build size because of the way they generate in JS" which is a non-issue here and only increases it by a few bytes (I guess unless you use a million of them, but we wouldn't be, and even then we're working with server side TypeScript so the "client downloads more data" still doesn't really apply either way) or "string enums don't allow string literals as input" but I wasn't proposing string enums to begin with, nor is that really an "issue" imo (just use the enum value? That's why it's there?)? I have not seen any super concrete reason as to why enums are so avoided here, and why people work around them so much?

This seems to be a take I exclusively see from the TypeScript community as well (despite things like https://github.com/microsoft/TypeScript itself using enums in thousands of places?), whereas enums are very embraced in other languages in my experience. We make use of them *all the time* in our Go NEX libraries for example and have had 0 actual issues here. The TypeScript docs don't have any warnings or anything about their usage either, and in my experience when docs contain features that shouldn't really be used anymore that warning is also documented?

I agree that magic numbers are bad as they are opaque without other context like comments, and comments shouldn't be a substitute for code, but I don't see why an enum is a worse option here especially since it means we have to keep doing these conversions? And even if enums really truly are something that should always be avoided, the functionality can still be emulated by just using plain objects:

```ts
const TokenType = {
  OAuthAccess: 0,
  OAuthRefresh: 1
} as const;

type TokenTypes = typeof TokenType[keyof typeof TokenType];

function generateToken(tokenType: TokenTypes) {
    console.log(tokenType);
}

generateToken(TokenType.OAuthAccess);
generateToken(TokenType.OAuthRefresh);

// generateToken(100); // Doesn't run, throws: Argument of type '100' is not assignable to parameter of type 'TokenType'.
// generateToken("0"); //  Doesn't run, throws: Argument of type '"0"' is not assignable to parameter of type 'TokenType'.
```

Which gets the same result as the enum

Though even when I Googled the phrase `why you should not use enums in typescript`, the 3 results were 1 dev.to article and then some Reddit threads. The Reddit threads being links to other articles/videos about why you should avoid enums, and the comments on these threads (which again are among the first results on Google, I did not need to look far), somewhat echo my confusion for the avoidance:

- https://www.reddit.com/r/typescript/comments/yr4vv5/typescript_enums_are_terrible_heres_why/
- https://www.reddit.com/r/javascript/comments/zsf541/avoid_using_enums_in_typescript/

With quotes such as:

> I really can't stand talks like this. Here we have a very opinionated person who is clearly missing a lot of information on the topic. The fact that he keeps saying "I don't know why" and "for some reason" means he's not really much of an expert and maybe shouldn't be presenting this information about enums being "TERRIBLE". I would recommend reviewing the documentation.
>
> String enums are transpiled down to plain JavaScript objects. Using as const and keyof / typeof syntax is a work around to create enumerations out of objects because it is inline with the JavaScript spec (see bottom of TS documentation on enums).
>
> If you want to convert your string enum into a string union, you can use this:
> ```ts
> enum AuthMethods {
>     push = 'Push',
>     sms = 'Sms',
>     voice = 'Voice'
> }
> 
> type AuthMethodStrings = `${AuthMethods}`
> // 
> // above is the same as:
> // type AuthMethodStrings = "Push" | "Sms" | "Voice"
> ```
> Enums exist in many compiled languages, like C++ and Java, and they provide a lot of utility in TypeScript as well. Maybe its not your preference and thats ok, but they are far from terrible.

And:

> tl;dr;
>
> Person had problems with enums, because by default they will return their index in enum
>
> person didn't like to specify string for enum, becuase then they had to use enum, not plain strings
>
> In the end they suggested fairly complex structure, just in order to be able to use plain strings as values and pass type checks

So it seems like even the "enums should always be avoided" sentiment in TypeScript is pretty well challenged?

> The conversion also acts as a validation step, ensuring that only valid values would be accepted

This would only be an issue if we are trying to consume tokens that we did not create, since we don't know the token contents at run time. If we didn't create a token, then it won't decrypt correctly to begin with. If the token decrypts, then that means we made the token, and we can easily ensure only valid values are used during creation which can equally be done with an enum:

```ts
enum TokenType {
    OAuthAccess,
    OAuthRefresh
}

function generateToken(tokenType: TokenType) {
    console.log(tokenType);
}

generateToken(TokenType.OAuthAccess);
generateToken(TokenType.OAuthRefresh);

// generateToken(100); // Doesn't run, throws: Argument of type '100' is not assignable to parameter of type 'TokenType'.
// generateToken("0"); //  Doesn't run, throws: Argument of type '"0"' is not assignable to parameter of type 'TokenType'.
```

Since we never have to deal with "user input" in the case of the token contents, trying to guard around it reads to me as something like this:

```js
const SOME_CONSTANT = 'hello';

if (SOME_CONSTANT !== 'hello') {
    throw new Error('SOME_CONSTANT not hello');
}
```

Which feels kinda silly to me?

src/services/grpc/api/login.ts
binaryoverload: Yeah in future I think making an issue on the common config repo would be better 👍

src/database.ts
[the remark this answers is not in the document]
binaryoverload: One of the reasons for using a string rather than an enum is that it's neater to have just the string rather than TokenType.AuthRefresh which requires almost twice as many characters and an import.


The problem with accepting a number type is that it's possible to not use the constant

When refactoring I found an instance where the API type was set to two different numbers - using a typed string eliminates this possibility 

---

Just tell me what you want changed and I'll do it, I just want to get this small PR merged and this dragging on because of a change that won't affect the end result is tiring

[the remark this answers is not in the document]
binaryoverload: I'm not sure I agree going number's is still the way to go, but in the interest of moving this PR along I've made the changes

PR #155, opened 2025-01-30
Account deletion/purging script

mrjvs:
Not an approved issue, but this is useful to have.

Based on the script made by Jon, slightly modified for better UX


binaryoverload (approved):
This has been used extensively, just adding it for team use

PR #157, opened 2025-01-31
Return the server a tier below if it does not exist

### LINKED ISSUE #150: [Enhancement]: For testers, when no testing server is available, return the production server
### Checked Existing

Currently, if a server set to testing environment is **not** available, users set to test servers will receive an error, causing us to need to duplicate servers across prod and test. However, it would be preferable if we didn't have to do this - if a test server is set this should override the production server for testing users, otherwise send them to the same production servers.

_No response_

AToska21:
Resolves #150 

### Changes:

If a server does not exist in a tier, use the tier below it.
For example, if a user is in the test access level, but a server for that game only exists on prod, use that server.
Same for the dev access level, but it goes dev --> test --> prod.

### DIFF
--- a/src/database.ts
+++ b/src/database.ts
@@ -22,6 +22,12 @@ const discordConnectionSchema = joi.object({
 	id: joi.string()
 });
 
+const accessModeOrder: Record<string, string[]> = {
+	prod: ['prod'],
+	test: ['test', 'prod'],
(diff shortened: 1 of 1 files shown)


AToska21:
It keeps setting the base branch to master, my bad.

PR #158, opened 2025-01-31
Return the server a tier below if it does not exist

### LINKED ISSUE #150: [Enhancement]: For testers, when no testing server is available, return the production server

AToska21:
Resolves #150

### Changes:

If a server does not exist in a tier, use the tier below it.
For example, if a user is in the test access level, but a server for that game only exists on prod, use that server.
Same for the dev access level, but it goes dev --> test --> prod.

This time I set the base branch to `dev`, ignore the old PR.

### DIFF
--- a/src/database.ts
+++ b/src/database.ts
@@ -212,14 +212,15 @@ export async function getPNIDProfileJSONByPID(pid: number): Promise<PNIDProfile 
 }
 
 export async function getServerByGameServerID(gameServerID: string, accessMode: string): Promise<HydratedServerDocument | null> {
-	const searchModes = accessModeOrder[accessMode] || ['prod'];
+	const searchModes = accessModeOrder[accessMode] ?? accessModeOrder.prod; // Default to prod if invalid mode
+
+	const servers = await Server.find({
+		game_server_id: gameServerID,
+		access_mode: { $in: searchModes }
+	})
 
 	for (const mode of searchModes) {
-		const server = await Server.findOne({
-			game_server_id: gameServerID,
-			access_mode: mode
-		});
-
+		const server = servers.find(s => s.access_mode === mode);
 		if (server) return server;
 	}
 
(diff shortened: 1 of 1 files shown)


src/database.ts
```diff
@@ -206,24 +212,48 @@ export async function getPNIDProfileJSONByPID(pid: number): Promise<PNIDProfile
 }
 
 export async function getServerByGameServerID(gameServerID: string, accessMode: string): Promise<HydratedServerDocument | null> {
-	return await Server.findOne({
-		game_server_id: gameServerID,
-		access_mode: accessMode
-	});
+	const searchModes = accessModeOrder[accessMode] || ['prod'];
```
binaryoverload, line 215: We want to use the `??` operator which communicates the intention clearly as it's specifically for null/undefined coalescing

We also want to use the `accessModeOrder.prod` so that the default is tied to the order object 

```suggestion
	const searchModes = accessModeOrder[accessMode] ?? accessModeOrder.prod;
```

src/database.ts
```diff
@@ -206,24 +212,48 @@ export async function getPNIDProfileJSONByPID(pid: number): Promise<PNIDProfile
...
+
+	for (const mode of searchModes) {
+		const server = await Server.findOne({
+			game_server_id: gameServerID,
+			access_mode: mode
+		});
+
+		if (server) return server;
+	}
```
binaryoverload, line 224: This makes potentially three database calls

What I think we should do instead is use `Server.find` and pass the `searchModes` to the `$in` operator

We can then do this loop to find the right server

binaryoverload (changes requested):

src/database.ts
AToska21: Fixed!

src/database.ts
AToska21: I think I might have fixed this? Not tested since I'm in a bit of a hurry right now.

AToska21: Can confirm it works.

binaryoverload (approved):
Looks good to me!

PR #165, opened 2025-03-22
Merge Dev to Master

binaryoverload:
A merge to master is long overdue, as we already run dev in production

The primary purpose of this PR is to get the Docker workflow onto master, so we can run the workflow dispatch action


DaniElectra:
The long awaited PR

PR #170, opened 2025-04-11
Update NNAS response/error handling

### LINKED ISSUE #160: [Enhancement]: Centralize response creation
### Checked Existing

Create a helper function to centralize the creation of responses in some services. This is mainly aimed at NNAS but might be useful in other services as well.

The motivation behind this is the fact that our response creation is very clunky in a lot of places. Lots of repeated code, some areas missing things, lots of duplicate calls to set headers, etc.

I noticed this when looking at how #111 adds in a warning comment for the OAuth endpoint. This comment is supposed to warn users that sharing details could compromise their data like their device certificate (we know of at least 1 case of a user being tricked into giving up their device certificate this way, which was the motivation for the comment). I wanted to add this warning to other endpoints, since basically every NNAS endpoint needs it, and realized how much duplicate code that would be.

I'm thinking maybe something like this?

```ts
import express from 'express';
import xmlbuilder from 'xmlbuilder';

type CreateNNASErrorResponseOptions = {
	status?: number;
	error: {
		cause?: string;
		code?: string;
		message?: string;
	}
}

type CreateNNASResponseOptions = {
	status?: number;
	body: Record<string, any>;
}

const router = express.Router();

router.post('/access_token/generate', async (request: express.Request, response: express.Response): Promise<void> => {
	// Call createNNASErrorResponse if error, otherwise createNNASResponse
});

function createNNASErrorResponse(response: express.Response, options: CreateNNASErrorResponseOptions): void {
	createNNASResponse(response, {
		status: options.status || 400,
		body: {
			error: options.error
		}
	});
}

function createNNASResponse(response: express.Response, options: CreateNNASResponseOptions): void {
	response.set('Content-Type', 'text/xml');
	response.set('Server', 'Nintendo 3DS (http)');
	response.set('X-Nintendo-Date', new Date().getTime().toString());

	const body = xmlbuilder
		.create(options.body)
		.commentBefore('WARNING! DO NOT SHARE ANYTHING IN THIS REQUEST OR RESPONSE WITH UNTRUSTED USERS! IT CAN BE USED TO IMPERSONATE YOU AND YOUR CONSOLE, POTENTIALLY GETTING YOU BANNED!').end();

	response.status(options.status || 200).send(body);
}
```

This way we can easily just call the helper functions and have them set the headers and warning and all that

It should be noted that I'm unsure how exactly to handle situations where a NNAS error expects a format like:

```xml
<error>
	<cause>grant_type</cause>
	<code>0004</code>
	<message>Invalid Grant Type</message>
</error>
```

Rather than:

```xml
<errors>
	<error>
		<code>0108</code>
		<message>Account has been banned</message>
	</error>
</errors>
```

It seems like sometimes the outer `<errors>` element is expected, and other times it is not. I could be mistaken about this though? @DaniElectra @shutterbug2000 @InternalLoss can we confirm if this is the case?

jonbarrow:
Resolves #160

### Changes:

Centralizes the response creation for NNAS. These changes automatically apply the common headers, as well as adds a comment to every response warning users not to share details with untrusted people. Also adds some notes/groundwork for supporting JSON mode in NNAS. Also fixes 3DS cert parsing as a bonus

Some of our error responses were formatted incorrectly, omitting the outer `<errors>` tag. This resulted in `102-2402` (`BAD_FORMAT_REQUEST`) since the client could not handle the error response. Now that this is fixed, we should get more accurate error codes from users

Marking as draft since it still needs robust testing on a console. I did some quick sanity checks using curl, but that's not enough

### DIFF
--- a/src/services/nnas/routes/oauth.ts
+++ b/src/services/nnas/routes/oauth.ts
@@ -1,9 +1,8 @@
 import express from 'express';
-import xmlbuilder from 'xmlbuilder';
 import bcrypt from 'bcrypt';
 import { getPNIDByNNASRefreshToken, getPNIDByUsername } from '@/database';
 import { generateOAuthTokens } from '@/util';
-import { createNNASErrorResponse } from '@/services/nnas/create-response';
+import { createNNASErrorResponse, createNNASResponse } from '@/services/nnas/create-response';
 import { Device } from '@/models/device';
 import { SystemType } from '@/types/common/token';
 
@@ -155,15 +154,17 @@ router.post('/access_token/generate', async (request: express.Request, response:
 	try {
 		const tokenGeneration = generateOAuthTokens(SystemType.WIIU, pnid);
 
-		response.send(xmlbuilder.create({
-			OAuth20: {
-				access_token: {
-					token: tokenGeneration.accessToken,
-					refresh_token: tokenGeneration.refreshToken,
-					expires_in: tokenGeneration.expiresInSecs.access
+		return createNNASResponse(response, {
+			body: {
+				OAuth20: {
+					access_token: {
+						token: tokenGeneration.accessToken,
+						refresh_token: tokenGeneration.refreshToken,
+						expires_in: tokenGeneration.expiresInSecs.access
+					}
 				}
 			}
-		}).commentBefore('WARNING! DO NOT SHARE ANYTHING IN THIS REQUEST OR RESPONSE WITH UNTRUSTED USERS! IT CAN BE USED TO IMPERSONATE YOU AND YOUR CONSOLE, POTENTIALLY GETTING YOU BANNED!').end()); // TODO - This is ugly
+		});
 	} catch {
 		response.status(500);
 	}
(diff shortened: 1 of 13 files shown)


src/services/nnas/routes/oauth.ts
```diff
@@ -20,110 +20,112 @@ router.post('/access_token/generate', async (request: express.Request, response:
...
+			return createNNASErrorResponse(response, {
+				status: 401,
+				errors: [
+					{
+						cause: 'refresh_token',
+						code: '0106',
+						message: 'Invalid Refresh Token'
+					}
+				]
+			});
```
mrjvs, line 85: All the `refresh_token` errors in this file have been changed from `400` to `401`.
Was that intentional? nowhere else you seem to have modified status codes.

jonbarrow: It was. Those are the status codes Nintendo sends, the other status codes were wrong

I was going to update ALL the errors across NNAS, but I still have a lot of research to do on that front and it's not critical to this PRs scope

mrjvs: Makes sense, was just making sure

PR #181, opened 2025-05-31
Remove WUP check when adding PNID to device in NNAS oauth

jonbarrow:
Resolves #XXX

### Changes:

For some reason we were checking for Wii U devices in `/v1/api/oauth20/access_token/generate` of NNAS, resulting in PNIDs not being properly added to 3DS devices. This removes that check.

Not pushing directly because the check seemed deliberate, but no one can seem to remember why it was added? This *shouldn't* break anything (as it would just make the 3DS have the same behavior as the Wii U?), but doing a PR just in case.

No official issue was made here but it was raised on Discord.

### DIFF
--- a/src/services/nnas/routes/oauth.ts
+++ b/src/services/nnas/routes/oauth.ts
@@ -130,15 +130,13 @@ router.post('/access_token/generate', deviceCertificateMiddleware, consoleStatus
 
 	// * This are set/validated in consoleStatusVerificationMiddleware
 	// * It is always set, despite what Express might think
-	if (request.device?.model === 'wup') {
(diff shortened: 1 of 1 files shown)


jonbarrow:
I agree that a better solution should be made. Would you want to see that in this PR though, or do you think that an issue should be made and handled in a later PR? As you said, a more robust version of this would require several other changes

The main reason for this PR as it stands is for the mods, since they have a hard time right now finding 3DS devices when all they have is a PNID (such as in Juxt), which is why this has been hacked into the oauth endpoint (so that accounts can be retroactively added to a 3DS), not necessarily a robust tracking/linking system

PR #188, opened 2025-06-21
feat(grpc): Add device linking data to account GetUserData(pid)

ashquarky:
Continued from https://github.com/PretendoNetwork/grpc/pull/3

### Changes:

Updates the account.v2 (backend) GetUserData API to add device linking data. See the above-linked PR for more rationale on this, but in short it will be used to properly enforce Juxt's "only accounts with linked consoles" rule.

I updated Juxt to use the newer v2 API and confirmed I was receiving correct device linking data.
image

This PR also splits the v1 and v2 implementations of the API (ta @jonbarrow)

- [x] I have read and agreed to the [Code of Conduct](https://github.com/PretendoNetwork/Pretendo/blob/master/.github/CODE_OF_CONDUCT.md).
- [x] I have read and complied with the [contributing guidelines](https://github.com/PretendoNetwork/Pretendo/blob/master/.github/CONTRIBUTING.md).
- [ ] What I'm implementing was an [approved issue](../issues?q=is%3Aopen+is%3Aissue+label%3Aapproved).
- [x] I have tested all of my changes.

PR #189, opened 2025-06-21
Update token refresh/expire times

### LINKED ISSUE #184: [Bug]: Service token expiry is too short
### Checked Existing

### What happened?

Tokens issued via NNAS currently have a 1 hour expiry time:
https://github.com/PretendoNetwork/account/blob/dev/src/services/nnas/routes/provider.ts#L92-L101
https://github.com/PretendoNetwork/account/blob/dev/src/services/nnas/routes/oauth.ts#L156-L171

This is too short for the Miiverse applets on Wii U and 3DS, which refresh their tokens after 12 hours. Presumably other applications are also made with this assumption in mind.

Tokens should be valid for at least 12 hours - my preference would be 24 just to have some buffer and ensure the user will never unexpectedly have a bad token.

### Steps to reproduce?

_No response_

This is blocking https://github.com/PretendoNetwork/juxtaposition/pull/114 and other authentication improvements (since we need to hack around token expiry using cookies). I would like to move Juxt towards much stronger validation of tokens since this is a commonly exploited area.

While it may seem that having long-lived tokens is bad for moderation, this would actually allow bans to apply immediately, since we could fearlessly validate the tokens over gRPC for every or almost every request. Right now, Juxt has to cache token validity for a very long time to try and work around this expiry problem, allowing users to bypass bans for a time.

The console applets do not use the refresh token mechanism as far as I'm aware. @DaniElectra can likely comment further on their exact behaviour.

jonbarrow:
Resolves #184 

### Changes:

Changes the token generation such that refresh tokens are now valid for 12 hours rather than valid for the same length as the access token (which made refresh tokens effectively useless), and made it so that service tokens only have an ISSUED time, not an EXPIRED time, so that independent services can expire tokens as they see fit

Marked as a draft because:

- Untested as of now
- Once https://github.com/PretendoNetwork/account/pull/188 merges this will need to be updated again
- The service token expire time ignoring is a *little* smelly. I want to make sure there's no auth holes before merging
- Maybe we should discuss moving ALL tokens to issued times rather than expire times?

CC @ashquarky @DaniElectra @shutterbug2000 for the above


src/database.ts
```diff
@@ -117,6 +119,12 @@ export async function getPNIDByTokenAuth(token: string, allowedTypes?: number[])
 
 		const pnid = await getPNIDByPID(unpackedToken.pid);
 
+		// TODO - This is a hack. If `allowedTypes` is not set, it is assumed to be a call from our internal API,
+		//        and we ignore expiration checks on service tokens. Independent services should expire their own tokens
+		if (unpackedToken.token_type === TokenType.IndependentService && !allowedTypes) {
+			return pnid;
+		}
+
```
ashquarky, line 127: I don't think this comment is accurate. I found the following callsites that could reach this condition:

# API auth
https://github.com/PretendoNetwork/account/blob/fc1963363adb6f0f9128a049057f6b138f02d726/src/middleware/api.ts#L5-L16

Looks like you could use a service token (which now won't ever expire) to directly authenticate to the API (/v1/user, /v1/reset-password, etc.)

# NNAS refresh tokens
From request.body:
https://github.com/PretendoNetwork/account/blob/fc1963363adb6f0f9128a049057f6b138f02d726/src/services/nnas/routes/oauth.ts#L20-L24
It gets passed right in:
https://github.com/PretendoNetwork/account/blob/fc1963363adb6f0f9128a049057f6b138f02d726/src/services/nnas/routes/oauth.ts#L93

So you could use this to convert (non-expiring) service tokens into valid OAuth creds by putting a service token in the refresh token field.

# API refresh tokens
Similar to above: 
https://github.com/PretendoNetwork/account/blob/fc1963363adb6f0f9128a049057f6b138f02d726/src/services/api/routes/v1/login.ts#L19-L23
https://github.com/PretendoNetwork/account/blob/fc1963363adb6f0f9128a049057f6b138f02d726/src/services/api/routes/v1/login.ts#L92

We could audit these cases and fix them, but I would argue service tokens should be validated with a totally seperate function so we can be more precise about *which APIs* accept these tokens. As far as I can tell, the only useful one for Juxt is ExchangeTokenForUserData? For pretty much everything else we go via the account API with our admin GRPC keys.

ashquarky: Third parties might also like if `GET /v1/user` took service tokens (+ some extra token metadata for them to make auth decisions with)

src/services/nasc/routes/ac.ts
```diff
@@ -89,12 +91,12 @@ async function processLoginRequest(server: HydratedServerDocument, pid: number,
...
 	const tokenOptions = {
-		system_type: 0x2, // * 3DS
-		token_type: 0x4, // * Service token
+		system_type: SystemType.CTR,
+		token_type: TokenType.IndependentService,
 		pid: pid,
 		access_level: 0,
 		title_id: BigInt(parseInt(titleID, 16)),
-		expire_time: BigInt(Date.now() + (3600 * 1000))
+		expire_time: BigInt(Date.now()) // TODO - Hack. Independent services expire their own tokens, so we give them the ISSUED time, not an EXPIRE time
```
ashquarky, line 99: I think it would be nice if service tokens had a different TS type. That's what we did on the Juxt end (since we never decrypt OAuth tokens there).

ashquarky (changes requested):
it's fucked! it's broken! :sob:

src/database.ts
jonbarrow: I knew this was smelly. To be clear when I said "internal" I was accounting for the API, since it's only officially used by our website on the backend, but I guess "internal" is a strong word there since it's exposed to the public and isn't protected really (also doesn't account for the NNAS ones, I just missed that entirely)

I agree with using a different function. That was something I had implemented before in a larger PR a while ago, but it was reverted by Will since something in that large PR completely broke logins somehow

We've slowly been moving the changes from that big PR over again one by one, so it might be a good time to do that here. You can see what I did here, I basically gave every single token type its own dedicated function https://github.com/PretendoNetwork/account/pull/111

src/services/nasc/routes/ac.ts
jonbarrow: With the impending token updates that will happen in the future, is it worth doing more than a hack?

This is somewhat related to a point I mentioned in the PRs opening comment though. Maybe we should just switch gears and give every token a "issued time" instead of an "expired time", for consistency. We would just need to change the expiration check a bit but that's not terribly hard

ashquarky: The issue with changing other token types is that there's an expiration check in nex-go, so we'd need to first update every NEX app that does token checks.

DaniElectra:
> I think I caught a bug in the NNAS stuff too - it was only ever checking for WUP and not allowing CTR

That's because the tokens generated with NNAS assume a WUP device

https://github.com/PretendoNetwork/account/blob/b9cf3194fc462a8d7e3e1a0b7953c084bc5ddb69/src/services/nnas/routes/oauth.ts#L156-L168

ashquarky:
> > I think I caught a bug in the NNAS stuff too - it was only ever checking for WUP and not allowing CTR
> 
> That's because the tokens generated with NNAS assume a WUP device

The code this is replacing explicitly allows CTR too, though? Is this just future-proofing?

https://github.com/PretendoNetwork/account/blob/2717e01febd712f66a40560345c15cbd9df11149/src/middleware/pnid.ts#L25-L28

DaniElectra:
Oh right, ideally we should set the proper system type on the token yeah, so that's probably future-proofing

jonbarrow:
> Oh right, ideally we should set the proper system type on the token yeah, so that's probably future-proofing

It's not future-proofing, that code is just an old hack/hot-patch I had made a while ago and I had made a mistake. It was intended to be replaced with real token validation, which is what I then added in #111 which only has a single WUP check [`302956d`](https://github.com/PretendoNetwork/account/pull/111/commits/302956d5f2508f96e3a2394db9a32caf728f0696#diff-724f334a2b47a5e7cbcab252ce04fc232071d0bc058c96ed6fd4ac96e5dff308R114-R117)

jonbarrow:
> Notably, there is now no way to get the account server to validate a service token - services must decrypt them locally. We should probably add an API for this for the benefit of third parties.

Right now we don't officially support 3rd parties at all so my concern is just with our official needs, and we can just use gRPC for this for ourselves

Not saying we shouldn't ever do this, but 3rd parties are just not a huge priority to me right now tbqh

src/database.ts
```diff
@@ -106,25 +106,22 @@ export async function getPNIDByBasicAuth(token: string): Promise<HydratedPNIDDoc
 	return pnid;
 }
 
-export async function getPNIDByTokenAuth(token: string, allowedTypes?: SystemType[]): Promise<HydratedPNIDDocument | null> {
+async function getPNIDByOAuthToken(token: string, expectedSystemTypes: SystemType[], expectedTokenType: TokenType): Promise<HydratedPNIDDocument | null> {
```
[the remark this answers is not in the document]
ashquarky: If you want? This isn't exported so I don't think it would be a big difference either way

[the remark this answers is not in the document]
jonbarrow: Once I see 3+ arguments I start to think of maybe those arguments would be better as an object, but if you don't think it's a big deal then that's fair enough 👍 not a huge deal either way

ashquarky:
So I'll remove CTR and then this should be good to go?

jonbarrow:
> So I'll remove CTR and then this should be good to go?

Should be, yeah

ashquarky (approved):

ashquarky:
@jonbarrow pls review and merge <3

DaniElectra (approved):

PR #192, opened 2025-07-30
feat(nasc): Add enforcement of custom patches to required titles

DaniElectra:
Resolves #XXX

### Changes:

If the supplied game server ID doesn't match the one present on the server database, then assume the title requires patches on the user's end. Give this behavior a custom error code 002-0152 to facilitate support to unaware users.

Also drop the legacy fallback of the friends server to the test environment, since the server should be available to anyone anyway, and we already have the production config for it.

Alongside that, update the maintenance error code to use the proper 002-0101 as expected.

### DIFF
--- a/src/middleware/nasc.ts
+++ b/src/middleware/nasc.ts
@@ -16,7 +16,8 @@ async function NASCMiddleware(request: express.Request, response: express.Respon
 		!requestParams.csnum ||
 		!requestParams.macadr ||
 		!requestParams.titleid ||
-		!requestParams.servertype
+		!requestParams.servertype ||
+		!requestParams.gameid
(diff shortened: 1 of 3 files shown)


jonbarrow (dismissed):
LGTM

jonbarrow:
Actually quick question, how will this work with future updates? For example say we find another exploit in ACNL, how will we enforce this? Is the idea that we just keep rotating the title ID? That's fine if so, just trying to make sure I'm on the same page there

DaniElectra:
I like that idea. At the moment I'll just leave it as a TODO though

jonbarrow (approved):
LGTM

PR #194, opened 2025-08-27
Mississippi

jonbarrow:
Resolves #XXX

### Changes:

Adds in a location/age check to block registration from users under the age of 18 who live in Mississippi. Mississippi recently passed a law that states that we cannot collect data for anyone under 18 without parental consent, but the ways the law provides to obtain consent are not viable to us. See https://bsky.social/about/blog/08-22-2025-mississippi-hb1126 for details

Rather than taking the nuclear option like Bluesky, we just block registration if the user is under 18 since we're still allowed to collect data for people in Mississippi above 18

This does not affect NASC or the game servers in general, since the law provides an exemption for video game servers. This mostly only affects Juxtaposition. But since you need a PNID to use Juxtaposition, just block the registration entirely and be done with it

The location database functionality was tested outside of this implementation and seems to work well, though IP2Location only allows you to download the databases 5 times per day, which I've used up, so I cannot test this further at this time

This also relies on some changes to the website. These changes have not happened yet as I wanted feedback on the overall design of this first

## It is important to note that we are not actively storing or tracking user locations. These are CHECKS, and the data is discarded immediately. We have no interest in collecting location data

Marking as a draft for now since the whole "IP2Location download limit" thing prevents further testing from me at this time

### DIFF
--- a/scripts/download-ip2location-databases.js
+++ b/scripts/download-ip2location-databases.js
@@ -39,7 +39,7 @@ async function main() {
 				}
 			}
 		} catch (error) {
-			console.error(error);
+			console.error('Error downloading IP2Location databases:', error);
 		}
 	}
 }


src/services/nnas/routes/people.ts
```diff
@@ -64,6 +65,33 @@ router.post('/', ratelimit, deviceCertificateMiddleware, async (request: express
 
 	const person: Person = request.body.person;
 
+	// TODO - This is kinda ugly
+	const birthdate = new Date(person.birth_date);
+	const today = new Date();
+	const eighteenthBirthday = new Date(birthdate);
+	eighteenthBirthday.setFullYear(birthdate.getFullYear() + 18);
```
binaryoverload, line 72: We should probably move this into a util function - I don't think it's ugly like the comment says - we can just put logic somewhere so we don't have to do it every time

Maybe a `getAgeFromDate(date)` function - which we can use smth like this:

```js
const age = getAgeFromDate(date);
if (age < 18) {
  // do things
}
```

Would allow for reuse with other age verification (like < 16, < 13)

---

Speaking of which, should we add a < 13 block to this code?

src/services/api/routes/v1/register.ts
```diff
@@ -68,6 +71,27 @@ router.post('/', async (request: express.Request, response: express.Response): P
 		}
 	}
 
+	// TODO - This is kinda ugly
+	const birthdate = new Date(birthday);
+	const today = new Date();
+	const eighteenthBirthday = new Date(birthdate);
+	eighteenthBirthday.setFullYear(birthdate.getFullYear() + 18);
```
binaryoverload, line 78: Same comment as below

scripts/download-ip2location-databases.js
```diff
@@ -0,0 +1,82 @@
+const { Readable } = require('node:stream');
+const fs = require('node:fs');
+const path = require('node:path');
+const unzipper = require('unzipper');
+
+require('dotenv').config();
+
+// * unzipper wants to use the "request" module, which is deprecated and insecure.
+// * Just wrap native fetch to avoid another dependancy here
+// TODO - This is kinda ugly, can this be better?
```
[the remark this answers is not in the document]
binaryoverload: https://github.com/PretendoNetwork/account/actions/runs/17278556458/job/49041474829?pr=194#step:6:384

I've also noticed that the logic for the `request` stuff is failing here - something to do with content-length - if we switch to yauzl and fetch I expect it'd be a non-issue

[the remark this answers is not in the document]
jonbarrow: Sounds good to me, I have 0 attachment to unzipper 👍 I just Googled for a lib to work with zip files in memory (most I saw required working with a physical file already on disk) and that was the first thing I saw. If yauzl can do that then works for me

src/services/nnas/routes/people.ts
```diff
@@ -64,6 +65,33 @@ router.post('/', ratelimit, deviceCertificateMiddleware, async (request: express
...
 
+	// TODO - This is kinda ugly
+	const birthdate = new Date(person.birth_date);
+	const today = new Date();
+	const eighteenthBirthday = new Date(birthdate);
+	eighteenthBirthday.setFullYear(birthdate.getFullYear() + 18);
+
+	if (today < eighteenthBirthday) {
+		// TODO - Enable `CF-IPCountry` in Cloudflare and only use IP2Location as a fallback
+		const ip = (request.headers['cf-connecting-ip'] || request.headers['x-forwarded-for'] || request.ip) as string | undefined;
```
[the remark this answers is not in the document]
jonbarrow: Works for me papa, will update when home 👍

src/services/nnas/routes/people.ts
jonbarrow: I agree with the util function. I think that will be useful for the other age checks like you said, but right now I'd like to keep this PR focussed on the Mississippi changes if possible

src/services/api/routes/v1/register.ts
```diff
@@ -37,6 +38,8 @@ const DEFAULT_MII_DATA = Buffer.from('AwAAQOlVognnx0GC2/uogAOzuI0n2QAAAEBEAGUAZg
  * Description: Creates a new user PNID
  */
 router.post('/', async (request: express.Request, response: express.Response): Promise<void> => {
+	const clientIP = request.body.ip?.trim(); // * This has to be forwarded since this request comes from the websites server
```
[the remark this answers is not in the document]
binaryoverload: If it's only for site to site, we should add a restriction so only the website can call this - afaik this can be called publically right now?

src/services/nnas/routes/people.ts
binaryoverload: So to clarify:
- You'll make the util function and implement it in this PR
- For the < 13 check, that will be for a different PR

Is that right?

src/services/api/routes/v1/register.ts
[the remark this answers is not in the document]
jonbarrow: Yes I believe it can be at the moment. I *THINK* someone mentioned this a while ago and it just got lost in the cracks? I believe we agreed to do what we do with gRPC and have a API key header that only the backend knows

Does that sound alright to you? Or would you prefer something different?

src/services/nnas/routes/people.ts
jonbarrow: Correct 👍

jonbarrow: To clarify, I want to do separate PRs for different laws/regions because they all vary in certain ways, and given how important the features are I want the PRs to be as focussed as possible on each law to make sure everything is nailed down correctly

src/services/api/routes/v1/register.ts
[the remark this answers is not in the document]
binaryoverload: API Key sounds perfectly fine to me 👍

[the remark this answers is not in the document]
binaryoverload: I'm happy to leave it out of this PR as long as it's done separately as a priority

scripts/download-ip2location-databases.js
```diff
@@ -70,12 +17,31 @@ const databases = {
...
-		const content = await file.buffer();
-		fs.writeFileSync(database.save_path, content);
+		const response = await fetch(`https://www.ip2location.com/download/?token=${process.env.PN_ACT_CONFIG_IP2LOCATION_TOKEN}&file=${name}`);
+		const arrayBuffer = await response.arrayBuffer();
+		const buffer = Buffer.from(arrayBuffer);
+		const zip = await yauzl.fromBuffer(buffer);
+
+		try {
+			for await (const entry of zip) {
+				console.log(entry.filename);
```
binaryoverload, line 34: Was this just a debug log? We should add some context to it if we intend to keep it - like "Processing zip file: X"

scripts/download-ip2location-databases.js
```diff
@@ -70,12 +17,31 @@ const databases = {
...
+				if (entry.filename === database.file_name) {
+					const readStream = await entry.openReadStream();
+					const writeStream = fs.createWriteStream(database.save_path);
+
+					await pipeline(readStream, writeStream);
+				}
+			}
+		} finally {
+			await zip.close();
```
binaryoverload, line 43: From the docs:

> fromBuffer(buffer, [options], [callback])
> ...
> If a ZipFile is acquired from this method, it will never emit the close event, and calling close() is not necessary.

So we can omit this I think?

scripts/download-ip2location-databases.js
```diff
@@ -70,12 +17,31 @@ const databases = {
...
+		const zip = await yauzl.fromBuffer(buffer);
+
+		try {
+			for await (const entry of zip) {
+				console.log(entry.filename);
+				if (entry.filename === database.file_name) {
+					const readStream = await entry.openReadStream();
+					const writeStream = fs.createWriteStream(database.save_path);
+
+					await pipeline(readStream, writeStream);
```
binaryoverload, line 39: God bless - thanks for using pipelines - means we avoid any issues

scripts/download-ip2location-databases.js
```diff
@@ -70,12 +17,31 @@ const databases = {
...
+				}
+			}
+		} finally {
```
binaryoverload, line 42: Worth adding a try-catch here?

src/util.ts
```diff
@@ -338,3 +338,49 @@ export function getValueFromHeaders(headers: IncomingHttpHeaders, key: string):
...
+	const currentYear = today.getFullYear();
+	const currentMonth = today.getMonth() + 1;
+	const currentDay = today.getDate();
+
+	let age = currentYear - birthYear;
+
+	// * Check if birthday has actually happened this year yet
+	if (currentMonth < birthMonth || (currentMonth === birthMonth && currentDay < birthDay)) {
+		age--;
+	}
```
binaryoverload, line 383: I think it might be better to do this with maths

```js
const yearMs = 1000 * 60 * 60 * 24 * 365;
const dateDiff = new Date() - new Date(currentYear, currentMonth, currentDay);
const age = Math.floor(dateDiff / yearMs);
```

binaryoverload (changes requested):
The changes look good - thanks

Just a few comments

scripts/download-ip2location-databases.js
jonbarrow: It was a leftover, my bad

scripts/download-ip2location-databases.js
jonbarrow: Done and done

scripts/download-ip2location-databases.js
```diff
@@ -0,0 +1,47 @@
...
+			for await (const entry of zip) {
+				if (entry.filename === database.file_name) {
+					const readStream = await entry.openReadStream();
+					const writeStream = fs.createWriteStream(database.save_path);
+
+					await pipeline(readStream, writeStream);
+				}
+			}
+		} catch (error) {
+			console.error(error);
```
binaryoverload, line 42: It would be useful to prefix the error with what the error is, in the case that the error isn't self-descriptive

i.e. `Error downloading IP2Location databases:`

binaryoverload (changes requested):
Looks good! Thanks for your efforts :P 

Just one last one (promise)

binaryoverload (approved):

PR #196, opened 2025-10-03
Update PNID scrubbing functionality and add account deletion gRPC endpoints

jonbarrow:
Resolves #XXX

### Changes:

Implements the account server portions of https://github.com/PretendoNetwork/grpc/pull/8:

- Updates `pnid.scrub` to fully delete the Stripe user automatically, if exists. This was previously done manually, with only the subscription being automatically deleted. Deleting the user will delete subscriptions though, so this is safe
- Updates `pnid.scrub` to anonymize forum user data automatically, if exists. This was previously done manually
- Updates `pnid.scrub` to remove Miiverse (Juxtaposition) user data automatically, if exists. This was previously done manually
- Adds `deleteAccount` methods to both the `account` and `api` gRPC endpoints
- Ran `npm audit fix` since there were a lot of outdated packages (there's still vulns being reported that couldn't be fixed automatically, but that's out of scope here)

Not fully tested since Juxtaposition lacks it's implementation of things. Individual parts, like the Discourse scrubbing, were tested in isolation however

### DIFF
--- a/SETUP.md
+++ b/SETUP.md
@@ -64,29 +64,32 @@ The Pretendo Network website uses this server as an API for querying user inform
 
 Configurations are loaded through environment variables. `.env` files are supported. All configuration options will be gone over, both required and optional. There also exists an example `.env` file
 
-| Name                                          | Description                                                                                    | Optional |
-| --------------------------------------------- | ---------------------------------------------------------------------------------------------- | -------- |
-| `PN_ACT_CONFIG_HTTP_PORT`                     | The HTTP port the server listens on                                                            | No       |
-| `PN_ACT_CONFIG_IP2LOCATION_TOKEN`             | Download token for https://lite.ip2location.com. Used to download the local IP databases       | Yes      |
-| `PN_ACT_CONFIG_MONGO_CONNECTION_STRING`       | MongoDB connection string                                                                      | No       |
-| `PN_ACT_CONFIG_MONGOOSE_CONNECT_OPTIONS_PATH` | Path to a `.json` file containing Mongoose connection options                                  | Yes      |
-| `PN_ACT_CONFIG_REDIS_URL`                     | Redis URL                                                                                      | Yes      |
-| `PN_ACT_CONFIG_EMAIL_SES_REGION`              | Amazon SES Region                                                                              | Yes      |
-| `PN_ACT_CONFIG_EMAIL_SES_ACCESS_KEY`          | Amazon SES Access Key                                                                          | Yes      |
-| `PN_ACT_CONFIG_EMAIL_SES_SECRET_KEY`          | Amazon SES Access Secret                                                                       | Yes      |
-| `PN_ACT_CONFIG_EMAIL_FROM`                    | Email "from" address                                                                           | Yes      |
-| `PN_ACT_CONFIG_S3_ENDPOINT`                   | s3 server endpoint                                                                             | Yes      |
-| `PN_ACT_CONFIG_S3_ACCESS_KEY`                 | s3 secret key                                                                                  | Yes      |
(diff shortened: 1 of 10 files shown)


SETUP.md
```diff
@@ -64,29 +64,35 @@ The Pretendo Network website uses this server as an API for querying user inform
...
+| `PN_ACT_CONFIG_DISCOURSE_FORUM_URL`           | Used for anonymizing user accounts on Discourse during account deletion. Forum anonymization skipped if not set | No       |
+| `PN_ACT_CONFIG_DISCOURSE_API_KEY`             | Used for anonymizing user accounts on Discourse during account deletion. Forum anonymization skipped if not set | No       |
+| `PN_ACT_CONFIG_DISCOURSE_API_USERNAME`        | Used for anonymizing user accounts on Discourse during account deletion. Forum anonymization skipped if not set | No       |
+| `PN_ACT_CONFIG_GRPC_MIIVERSE_HOST`            | Used to remove Miiverse user data during account deletion                                                       | Yes      |
+| `PN_ACT_CONFIG_GRPC_MIIVERSE_PORT`            | Used to remove Miiverse user data during account deletion                                                       | Yes      |
+| `PN_ACT_CONFIG_GRPC_MIIVERSE_KEY_API`         | Used to remove Miiverse user data during account deletion                                                       | Yes      |
```
mrjvs, line 98: Seems like the "optional" values are the other way around?
miiverse is not optional and its marked as optional, and forum is optional and marked as not optional.

src/models/pnid.ts
mrjvs, line 1: Something that's still missing from pnid.scrub which was part of the manual deletion process is the deleting of mii CDN assets.

I forgot to mention it last time I talked it about it with you, sorry.

Can that still be added to this PR? If not, then I can probably add it to the juxt side.

src/models/pnid.ts
```diff
@@ -223,18 +229,53 @@ PNIDSchema.method('scrub', async function scrub() {
...
-				LOG_WARN(`SCRUBBING USER DATA FOR USER ${this.username}. HAS STRIPE SUBSCRIPTION ${this.connections.stripe.subscription_id}, BUT STRIPE CLIENT NOT ENABLED. SUBSCRIPTION NOT CANCELED`);
+				LOG_WARN(`SCRUBBING USER DATA FOR USER ${this.username}. HAS STRIPE DATA UDER ID ${this.connections.stripe.customer_id}, BUT STRIPE CLIENT NOT ENABLED.`);
 			}
 		} catch (error) {
-			LOG_ERROR(`ERROR REMOVING ${this.username} STRIPE SUBSCRIPTION. ${error}`);
+			LOG_ERROR(`ERROR REMOVING ${this.username} STRIPE DATA. ${error}`);
+		}
+	}
+
+	if (config.discourse.forum_url && config.discourse.api_key && config.discourse.api_username) {
```
mrjvs, line 251: There's error handling for miiverse (log and continue) but not for forum. Is that intentional?

mrjvs (changes requested):
Looks good, thanks for adding it this quickly.

SETUP.md
jonbarrow: You're right, oops

src/models/pnid.ts
jonbarrow: Since it's just using fetch I didn't think there was a path where this would throw anything, is that not the case? Stripe will throw on certain errors rather than just returning which is why there's error handling there

mrjvs: Fetch can throw errors on network issues or if the body is not json parsable (like if cloudflare shows a gateway timeout error or smth)

jonbarrow: Oh right I didn't even think about that

binaryoverload (approved):

mrjvs (approved):

PR #308, opened 2025-12-15
Add NASC `uidhmac` repair

jonbarrow:
Resolves #XXX

### Changes:

Adds the ability to remake the users `uidhmac` when needed. Untested, but should work.

@DaniElectra before we decide to merge this I wanted to talk about the actual hmac itself. Nintendo seemed to always use an 8 hex character (4 byte) hmac here. But that seems...not great. 4 bytes isn't a ton. When discussing this with Kinnay and Zak a while ago, I apparently was able to get the 3DS to use a larger string (https://github.com/kinnay/NintendoClients/pull/76#issuecomment-1017527962). But I have not tried this in ages and I'm not sure if maybe I just did something funky?

Can you confirm whether or not we're able to use larger strings here? I'm asking because I've checked some NASC dumps from our users and they seem to only be using the first 4 bytes, despite the fact that [our friends server seems to be sending the entire hmac](https://github.com/PretendoNetwork/friends/blob/bba53eb3ef3433d58c11de26bdaaeb7098b70456/nex/account-management/nintendo_create_account.go)? Which doesn't line up with what I apparently saw happen in 2022, maybe Nintendo changed something in one of the later updates...? I have not looked into the internals of the NASC client, but I figured you might know. If so, that would be ideal. If not, maybe we can use more than just hex characters?


jonbarrow:
@DaniElectra poke

DaniElectra (changes requested):
The limit for the PID HMAC is [8 UTF-16 characters](https://www.3dbrew.org/wiki/FRD_Savegame#account), so while we have some limitations there is some room for improving the original algorithm here.

Also to keep in mind, we have to change the PID HMAC generation/assignment on the friends server too for new accounts.

jonbarrow:
> (at least on the NASC login phase, unsure on account registration)

I asked ZeroSkill to take a look at this part since he was working with the 3DS NASC client a while ago. I confirmed that the only limitation is the length of the string, the console just copies from the buffer into the save directly. So this SHOULD be good!

Won't merge yet though because of the holiday and because the friends server needs to update as well, and to ask around to make sure the algorithm has no security issues

src/services/api/routes/v1/repair-uidhmac.ts
```diff
@@ -0,0 +1,77 @@
...
+	}
+
+	try {
+		const nexAccount = await NEXAccount.findOne({
+			pid: parseInt(pid),
+			password: nexPassword
+		});
+
+		if (!nexAccount) {
+			response.json({
```
DaniElectra: Shouldn't this be `response.status(400).json`?

jonbarrow: Sorry I'm just now seeing this, yes

jonbarrow: Fixed

PR #312, opened 2026-01-01
Server provisioning

mrjvs:
Changes:
- Added configuration setting to specific a server provisioning file. It will be used on boot once and after detected file changes.
- Add provisioning system. It will update existing server DB entries and update them with info from the provisioning file.

I added this system so I don't have to manually update IPs and ports with every infra change. I've had to do it at least 20 times last week in infra work (and I made a lot of mistakes, causing downtime). With this PR, it can be fully automated (also less mistakes due to less manual work)

### DIFF
--- a/src/config-manager.ts
+++ b/src/config-manager.ts
@@ -13,7 +13,8 @@ export const disabledFeatures = {
 	email: false,
 	captcha: false,
 	s3: false,
-	datastore: false
+	datastore: false,
+	serverProvisioning: false
 };
 
 const hexadecimalStringRegex = /^[0-9a-f]+$/i;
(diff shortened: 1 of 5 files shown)


src/config-manager.ts
```diff
@@ -282,6 +286,11 @@ if (!config.datastore.signature_secret) {
 	}
 }
 
+if (!config.provisioning.server_config) {
+	LOG_WARN('A server provisioning config file as not been set. Disabling feature, no server data will be provisioned. To enable feature set the PN_ACT_PROVISIONING_SERVER_CONFIG environment variable');
+	disabledFeatures.serverProvisioning = true;
+}
```
binaryoverload, line 292: Would it be worth checking if the file exists here? Given that this is validation

binaryoverload (changes requested):

binaryoverload (approved):

src/config-manager.ts
mrjvs: I think the provisioning workflow should handle it, it's not a "config invalid" case

Issue #108, opened 2024-08-09
[Enhancement]: Make tokens use RSA again


jonbarrow:
### Checked Existing

- [X] I have checked the repository for duplicate issues.

### What enhancement would you like to see?

Migrate tokens back RSA. Years ago we used RSA encryption for tokens, but moved away from it to AES for simplicity and to make the tokens smaller. This is not super great however, since it prevents us from securely allowing 3rd party servers and if the AES keys ever leaked anyone could generate their own.

### Any other details to share? (OPTIONAL)

Nintendo seems to have likely used RSA-1024 based on the size of the Wii U NEX token. However a few things should be considered:

1. It looks like ***access and refresh*** tokens do not appear to use RSA. They are very short, hex-encoded, strings which tells me they likely still use AES and/or some randomization? Access keys are always 16 bytes long, which screams AES, but refresh tokens are always 20 bytes long, which screams random? Since these tokens are not used by 3rd parties at all, only within the NNAS server, these can probably stay as AES or even entirely randomized and use https://github.com/PretendoNetwork/account/issues/107 to track their data
2. 3DS tokens from NASC ***might*** not be entirely RSA either. They always seem to be 105 bytes long, which could mean they are AES still? The guys from Colors! 3D made mention that they're able to decrypt the tokens to get basic data but then they ask a Nintendo server to verify a token signature, so maybe these use a combination of RSA+AES? Unsure, did not ask and I likely won't

jonbarrow:
I've done some tests with tokens, and these are the results:

- Access tokens from `/v1/api/oauth20/access_token/generate` will always be capped to 32 characters. Sending anything longer will be truncated to 32 characters
- Given than access tokens are capped at their official value, we can safely assume refresh tokens are also capped at their official value of 40 characters
- Both NEX and service tokens from NNAS will always be capped to 512 characters. Sending anything longer will be truncated to 512 characters
- Both NEX and service tokens from NASC will always be capped to 255 characters. Sending anything longer will be truncated to 255 characters

These are the ranges we have to work with here

PR #313, opened 2026-01-02
Opaque tokens

### LINKED ISSUE #107: [Enhancement]: Store tokens in database
### Checked Existing

- [X] I have checked the repository for duplicate issues.

Store tokens in the database. This is not an uncommon practice to do, and would allow us much more flexibility. @SuperMarioDaBom has suggested this in the past, so I'm officially adding it here.

By storing the tokens in a database, we can easily implement a token revoking system. This would allow us to create a "logout from all devices" feature, or revoke specific tokens after certain actions (such as account deletion). It would also allow us to store MUCH more data about the token. Right now tokens are encrypted, and try to store as much information as they can. However both the Wii U and 3DS have a size limit on tokens, so we cannot store ALL data in the tokens (making systems like JWT not viable, this was actually the first system we tried to use). By storing data about the tokens in the database with the token we can store as much data as we want and use the token as a lookup for it.

We would still want to keep the token body encrypted and contain *some* data. There are cases where this is useful to have, such as player tracking in environments where we may not actually care about the players data/account status. We also should likely not store the token as-is in the database, to prevent issues in case that collection is somehow breached. Storing the hashes of tokens should be fine though.

By storing them in the database we can also make use of the [Mongo TTL](https://www.mongodb.com/docs/manual/tutorial/expire-data) feature to make tokens automatically expire. Though we may not want this, since this data could be useful for tracking login statistics? Although statistics like that can also be better handled in something dedicated.

jonbarrow:
Resolves #107 and #108

### Changes:

Moves tokens to random data instead of storing the data directly. Trying to fit all the data inside the token was a losing battle because of how space-limited it all ways. This removes clients abilities to handle data entirely locally, but that's fine since it's only a single network hop. I made the tokens be the length of what they are from the official sources, despite the fact that some can sometimes be bigger, just to keep things simple

Marking as a draft because this isn't finished. The entire gRPC side needs to be done, which relies on https://github.com/PretendoNetwork/grpc/pull/10 being merged. Also for gRPC, I think we should just ditch the v1 implementation entirely since it's token handling is fundementally incomaptible with this new handling. Just cut our losses there


src/services/nnas/routes/oauth.ts
```diff
@@ -167,33 +167,44 @@ router.post('/access_token/generate', deviceCertificateMiddleware, consoleStatus
...
+		token: crypto.randomBytes(20).toString('hex'),
+		client_id: clientID,
+		client_secret: clientSecret,
+		pid: pnid.pid,
+		info: {
+			system_type: SystemType.WUP,
+			token_type: TokenType.OAuthRefresh,
+			title_id: BigInt(0), // TODO - Add this?
+			issued: new Date(),
+			expires: new Date(Date.now() + 12 * 3600 * 1000)
```
mrjvs: Refresh tokens and access tokens should not expire at the same time. (iirc, this was implemented as a fix sometime before)

src/services/api/routes/v1/register.ts
```diff
@@ -419,32 +420,34 @@ router.post('/', async (request: express.Request, response: express.Response): P
...
-		const refreshToken = refreshTokenBuffer ? refreshTokenBuffer.toString('hex') : '';
+		const accessToken = await OAuthToken.create({
+			token: crypto.randomBytes(16).toString('hex'),
+			client_id: 'a2efa818a34fa16b8afbc8a74eba3eda', // TODO - This is the Wii U config, change this?
+			client_secret: 'c91cdb5658bd4954ade78533a339cf9a', // TODO - This is the Wii U config, change this?
+			pid: pnid.pid,
+			info: {
+				system_type: SystemType.API,
+				token_type: TokenType.OAuthAccess,
+				title_id: BigInt(0),
```
mrjvs, line 451: Same as my other comment: access and refresh should not have the same expiry

src/services/api/routes/v1/login.ts
```diff
@@ -112,30 +113,34 @@ router.post('/', async (request: express.Request, response: express.Response): P
...
+			client_id: 'a2efa818a34fa16b8afbc8a74eba3eda', // TODO - This is the Wii U config, change this?
+			client_secret: 'c91cdb5658bd4954ade78533a339cf9a', // TODO - This is the Wii U config, change this?
+			pid: pnid.pid,
+			info: {
+				system_type: SystemType.API,
+				token_type: TokenType.OAuthAccess,
+				title_id: BigInt(0),
+				issued: new Date(),
+				expires: new Date(Date.now() + 12 * 3600 * 1000)
+			}
```
mrjvs, line 131: Same as my other comment: access and refresh should not have the same expiry

src/models/independent-service-token.ts
```diff
@@ -0,0 +1,41 @@
...
+	token: string;
+	client_id: string;
+	title_ids: string[];
+	pid: number;
+	info: {
+		system_type: number;
+		token_type: number;
+		title_id: bigint;
+		issued: Date;
+		expires: Date;
```
mrjvs, line 15: It was mentioned in a couple comments that IndependentServiceTokens dont expire.

In that case, it should be made in a way that doesn't keep generating new tokens - itll put a strain on our infra:
Only create a new token if there isn't already one for that pid, client_id and title_ids.
Otherwise, just return the existing token.

Alternatively, you can add a real expiry thats a lot longer, like a month. We just need to make sure no games require it for longer than a month.

src/models/oauth-token.ts
```diff
@@ -0,0 +1,41 @@
...
+	client_secret: string;
+	pid: number;
+	info: {
+		system_type: number;
+		token_type: number;
+		title_id: bigint;
+		issued: Date;
+		expires: Date;
+	};
+}
```
mrjvs, line 17: OAuthTokens, NexTokens and PasswordResetTokens all have an expiry. The tokens should be removed from the database once they expire so they don't put a strain on the infra.

I recommend doing a daily clear of old tokens from the database. We have a similar system implemented in BOSS: https://github.com/PretendoNetwork/BOSS/blob/dev/src/scheduled.ts

It can be as simple as doing 3 quick DB calls like this:

```ts
await OAuthToken.deleteMany({
	"info.expires": {
		$lt: new Date()
	}
});
```

mrjvs (dismissed):

mrjvs:
I haven't checked token sizes in my review, will possibly need to be checked by someone else

src/models/oauth-token.ts
jonbarrow: I was thinking of using mongos built in TTL feature for this but iirc @DaniElectra had reasons for keeping the data, I've forgotten now though. Dani can you confirm or deny?

src/models/independent-service-token.ts
jonbarrow: Service tokens never expire officially, and some clients hold onto them indefinitely. For example the WWP client on the Wii U seems to request the service token only a single time and caches it forever. We ignore the service token in WWP so that doesn't matter to us, but it's a sign that these tokens never expired and so we can't do so either just in case

The idea is that individual services can decide when a token expires when it needs to expire it, tokens themselves will always work on our end. The issue with returning the same token every time is that it prevents the service from being able to request a new token once it's decided a token should no longer be valid for their service, since they'd just keep getting the same token over and over with the same issuing date

I considered deleting existing tokens when requesting a new one, but clients are allowed to have multiple tokens at the same time. For example WWP again, that's assigned to the same client ID as the Miiverse API games uses iirc, so we can't just restrict clients to a single token per user

mrjvs: It's pretty unusual to have infinite lifetime tokens, even for Nintendo. I would guess that WWP adopts the strategy of "try to use it, request new token if it fails", but we'll have to test that to find out.

However with the constraints of infinite lifetime token, it does seem like we'll need to make a stateless token for service tokens specifically. It's a shame, but we can make it easier and more secure than before:

```ts
type ServiceTokenOptions = {
  titleId: string
  pid: number,
}

const serviceTokenMagic = 0x82;
const serviceTokenSize = 36;

function createServiceToken(ops: ServiceTokenOptions): string {
  const buffer = Buffer.alloc(serviceTokenSize);
  buffer.writeUInt8(serviceTokenMagic, 0);
  buffer.writeUInt32BE(ops.pid, 1);
  buffer.writeBigUInt64BE(BigInt(parseInt(ops.titleId, 16)), 5);
  buffer.writeBigUInt64BE(BigInt(Date.now()), 13);
  const dataSize = 21;

  const hmac = crypto
    .createHmac("sha256", insertSecretHere)
    .update(buffer.subarray(0, dataSize))
    .digest()
    .subarray(0, serviceTokenSize - dataSize);
  hmac.copy(buffer, dataSize);

  return buffer.toString('base64');
}
```

The same token can then be decrypted and verified in the GRPC methods. We just need to make sure that we don't introduce code in other services to pull data out of it. We'd lose the flexibility to change it later.

The actual data from the token is unsecured with this method, but I dont think the pid, creation time and title id is an issue to keep unsecured. The token is safe from tampering though, which is the important part.

jonbarrow: Actually, it looks like I was wrong about that. Unfortunately, Nintendo shut down their services like Miiverse (and thus WWP) prior to me working on things, so I didn't have live data to work with, so I had to make some assumptions based on the behavior I could still see. In all my dumps that have WWP data, I can see the same service token reused indefinitely, which is what led me to believe the tokens never expired

It seems like Nintendo just did something funky here, or just ignored the expiration date entirely on their end specifically for WWP? After reading your comment I also thought it didn't sound quite right that they just NEVER expire, so I checked my DMs I had with the developer of Colors! 3D, since we had to discuss the apps token needs so we could add official support for them. In those DMs he said:

> The token has some general info, like user account, application ID, expiration date and stuff. But nothing we use

So looks like I was wrong here, the tokens did have an expiration date and Nintendo probably just didn't care about it for WWP

That being said:

> I would guess that WWP adopts the strategy of "try to use it, request new token if it fails"

There is no known error that triggers a client to request a new token. The Colors! 3D dev did vaguely mention some server-side validation of tokens, so there's clearly more to it, but I didn't ask for details for obvious reasons. My guess is that Nintendo just used the state from the token in WWP to do whatever in their WWP API and just ignored the expiration?

We can probably employ a similar strategy here though:

- Keep service tokens stateful like you suggested
- Store them in the database still with a proper expiration. Say 24 hours maybe? WWP is the only known client that stores these tokens long term, and it's a special case clearly, so I don't think we need to worry about letting these tokens be long-lived anymore (all other clients do seem to request new ones during their session, or only for one-time actions like WATCH_DOGS linking your NNID to your UPlay account)
- Expire the tokens like the other tokens, but since there's still basic state a client like WWP can choose to ignore the expiration if it wants to
- Add a new gRPC method `ValidateIndependentServiceToken` which simply verifies the hmac is correct (even if the token is expired), so clients can keep using a token post expiration should they want to

src/services/nnas/routes/oauth.ts
jonbarrow: My bad, this was a bad copy-paste job

src/models/oauth-token.ts
jonbarrow: I gave each token collection a TTL index

src/models/independent-service-token.ts
jonbarrow: Added as described above

src/models/oauth-token.ts
DaniElectra: If we set a TTL to the tokens wouldn't that make WWP not be able to validate the token since the HMAC won't be on the server anymore?

There was this other suggestion to renovate the tokens when the client requests a new one, which I agree with

jonbarrow: WWP already doesn't check the token, it ignores the token last I checked because of the aforementioned token caching and we've changed the token format over the years (so peoples old cached tokens wouldn't be able to be validated anyway)

That being said, even if it does begin to validate tokens WWP uses service tokens which have the data still in token body itself. The gRPC PR supports validating service tokens which have expired since the HMAC data is present in the token, regardless of if there's a DB entry or not since all the server needs to do is check the HMAC. The client can then just pull the data from the token body, we could really remove the DB entirely for service tokens tbh but I kept it for consistency

DaniElectra: Oh I'm dumb I thought we were just sending the HMAC to the clients (haven't looked deeply into this PR yet lmao). Disregard this then

jonbarrow:
This should be ready for review now. I did test that it builds locally and it does, though not much else. It should all work fine though, the changes just replace manual token handling with DB lookups so nothing huge really

I'm not going to be here tomorrow however, so do feel free to make any required bug fixes/changes while I'm out just to push things along

src/services/api/routes/v1/login.ts
```diff
@@ -112,30 +113,37 @@ router.post('/', async (request: express.Request, response: express.Response): P
...
+				expires: new Date(Date.now() + (3600 * 1000))
+			}
+		});
 
-		const accessToken = accessTokenBuffer ? accessTokenBuffer.toString('hex') : '';
-		const newRefreshToken = refreshTokenBuffer ? refreshTokenBuffer.toString('hex') : '';
+		await OAuthToken.create({
+			token: crypto.createHash('sha256').update(newRefreshToken).digest('hex'),
+			client_id: 'a2efa818a34fa16b8afbc8a74eba3eda', // TODO - This is the Wii U config, change this?
+			client_secret: 'c91cdb5658bd4954ade78533a339cf9a', // TODO - This is the Wii U config, change this?
```
DaniElectra, line 137: Couldn't we just set these to the value that is set on the request headers?

src/services/grpc/account/v2/exchange-oauth-token-for-user-data.ts
```diff
@@ -0,0 +1,124 @@
...
+	try {
+		oAuthToken = await OAuthToken.findOne({
+			token: crypto.createHash('sha256').update(request.token).digest('hex')
+		});
+
+		if (!oAuthToken) {
+			throw new ServerError(Status.INVALID_ARGUMENT, 'Invalid token');
+		}
+
+		// TODO - Add checks for the client ID/title ID and matching system/token types here
```
DaniElectra, line 27: This comment is inaccurate, there isn't any title ID in use here, this should be about client ID / client secret

src/services/api/routes/v1/login.ts
jonbarrow: This is in the `api` service, which isn't used by consoles. This was originally designed for the website before we started the move to gRPC for everything, and iirc there's still some legacy applications using this endpoint so it can't be entirely removed yet

Since it's used by stuff like the website, there is no client ID/secret headers, so I just default them to the Wii U for now

Once everything moves to gRPC though and add proper OAuth application support, we can add this via gRPC headers

DaniElectra: Gotcha :+1:

jonbarrow:
This should be ready for review again. It had some merge conflicts which I believe I got resolved correctly? Tagging in @DaniElectra, @mrjvs and @binaryoverload though to check over my tracks and make sure I didn't break anything

src/services/api/routes/v1/resetPassword.ts
binaryoverload, line 177: Once the password reset is successful, the password reset token should be deleted

src/services/grpc/account/v2/exchange-independent-service-token-for-user-data.ts
```diff
@@ -0,0 +1,153 @@
...
+				pid: nexAccount.pid,
+				owningPid: nexAccount.owning_pid,
+				accessLevel: nexAccount.access_level,
+				serverAccessLevel: nexAccount.server_access_level,
+				friendCode: nexAccount.friend_code,
+				deviceType: nexAccount.device_type
+			},
+			tokenInfo: {
+				systemType: serviceToken.info.system_type as any, // TODO - Stop the any usage
+				tokenType: serviceToken.info.system_type as any, // TODO - Stop the any usage
```
binaryoverload: That should be token_type?

src/services/grpc/account/v2/exchange-nex-token-for-user-data.ts
```diff
@@ -0,0 +1,90 @@
...
+				pid: nexAccount.pid,
+				owningPid: nexAccount.owning_pid,
+				accessLevel: nexAccount.access_level,
+				serverAccessLevel: nexAccount.server_access_level,
+				friendCode: nexAccount.friend_code,
+				deviceType: nexAccount.device_type
+			},
+			tokenInfo: {
+				systemType: nexToken.info.system_type as any, // TODO - Stop the any usage
+				tokenType: nexToken.info.system_type as any, // TODO - Stop the any usage
```
binaryoverload: That should be token_type?

src/services/grpc/account/v2/exchange-oauth-token-for-user-data.ts
```diff
@@ -0,0 +1,124 @@
...
+			pid: nexAccount.pid,
+			owningPid: nexAccount.owning_pid,
+			accessLevel: nexAccount.access_level,
+			serverAccessLevel: nexAccount.server_access_level,
+			friendCode: nexAccount.friend_code,
+			deviceType: nexAccount.device_type
+		},
+		tokenInfo: {
+			systemType: oAuthToken.info.system_type as any, // TODO - Stop the any usage
+			tokenType: oAuthToken.info.system_type as any, // TODO - Stop the any usage
```
binaryoverload: That should be token_type?

src/services/grpc/account/v2/exchange-password-reset-token-for-user-data.ts
```diff
@@ -0,0 +1,120 @@
...
+				uploadBossFiles: pnid.hasPermission(PNID_PERMISSION_FLAGS.UPLOAD_BOSS_FILES),
+				updateBossFiles: pnid.hasPermission(PNID_PERMISSION_FLAGS.UPDATE_BOSS_FILES),
+				deleteBossFiles: pnid.hasPermission(PNID_PERMISSION_FLAGS.DELETE_BOSS_FILES),
+				updatePnidPermissions: pnid.hasPermission(PNID_PERMISSION_FLAGS.UPDATE_PNID_PERMISSIONS)
+			},
+			linkedDevices: devices
+		},
+		tokenInfo: {
+			systemType: passwordResetToken.info.system_type as any, // TODO - Stop the any usage
+			tokenType: passwordResetToken.info.system_type as any, // TODO - Stop the any usage
```
binaryoverload: That should be token_type?

src/services/nnas/routes/provider.ts
```diff
@@ -90,25 +93,32 @@ router.get('/service_token/@me', async (request: express.Request, response: expr
...
-		expire_time: BigInt(Date.now()) // TODO - Hack. Independent services expire their own tokens, so we give them the ISSUED time, not an EXPIRE time
+		title_id: titleID,
+		issued: new Date(),
+		expires: new Date(Date.now() + 24 * 3600 * 1000)
 	};
 
-	const serviceTokenBuffer = await generateToken(server.aes_key, tokenOptions);
-	let serviceToken = serviceTokenBuffer ? serviceTokenBuffer.toString('base64') : '';
-
-	if (request.isCemu) {
```
binaryoverload, line 105: This cemu logic has been removed - is that correct?

src/models/nex-token.ts
```diff
@@ -0,0 +1,41 @@
...
+	info: {
+		system_type: Number,
+		token_type: Number,
+		title_id: BigInt,
+		issued: Date,
+		expires: Date
+	}
+});
+
+NEXTokenSchema.index({ 'info.expires': 1 }, { expireAfterSeconds: 0 });
```
binaryoverload, line 39: Please add an index for the token too

src/models/independent-service-token.ts
```diff
@@ -0,0 +1,43 @@
...
+	info: {
+		system_type: Number,
+		token_type: Number,
+		title_id: BigInt,
+		issued: Date,
+		expires: Date
+	}
+});
+
+IndependentServiceTokenSchema.index({ 'info.expires': 1 }, { expireAfterSeconds: 0 });
```
binaryoverload, line 41: Please add an index for the token too

src/models/oauth-token.ts
```diff
@@ -0,0 +1,43 @@
...
+	info: {
+		system_type: Number,
+		token_type: Number,
+		title_id: BigInt,
+		issued: Date,
+		expires: Date
+	}
+});
+
+OAuthTokenSchema.index({ 'info.expires': 1 }, { expireAfterSeconds: 0 });
```
binaryoverload, line 41: Please add an index for the token too

src/models/password-reset-token.ts
```diff
@@ -0,0 +1,39 @@
...
+	info: {
+		system_type: Number,
+		token_type: Number,
+		title_id: BigInt,
+		issued: Date,
+		expires: Date
+	}
+});
+
+PasswordResetTokenSchema.index({ 'info.expires': 1 }, { expireAfterSeconds: 0 });
```
binaryoverload, line 37: Please add an index for the token too

binaryoverload (changes requested):
Some things I spotted

src/services/nnas/routes/provider.ts
jonbarrow: Yes. This logic was a hold-over from before Cemu got official support for our servers and is slowly being phased out. This was required on Cemu v1 because Cemu used to fail to parse our tokens correctly due to it only being coded to handle hex strings, but we were using base64 tokens. The `isCemu` flag is only set if the domain being used is `c.account.pretendo.cc` (a hold-over from how the old CemuHook patches worked, because Cemu was not open source yet we needed to use custom patches), which Cemu v2+ does not use and we no longer support v1 officially

binaryoverload (approved):
Looks good to me

src/services/nasc/routes/ac.ts
```diff
@@ -61,60 +64,72 @@ router.post('/', async (request: express.Request, response: express.Response): P
...
-		pid: pid,
-		access_level: 0,
-		title_id: BigInt(parseInt(titleID, 16)),
-		expire_time: BigInt(Date.now() + (3600 * 1000))
-	};
-
-	// TODO - Handle null tokens
+async function processLoginRequest(server: HydratedServerDocument, pid: number, requestParams: NASCLoginACRequestParams): Promise<URLSearchParams> {
+	const titleID = nintendoBase64Decode(requestParams.titleid).toString();
+	const token = nintendoBase64Encode(crypto.randomBytes(112));
```
DaniElectra, line 79: Is the intent here to make the token hash using the Nintendo-encoded base64, or should it use the regular base64 for the hash? Asking since the NEX side currently converts the Nintendo base64 to normal base64 (inherited from the previous token implementation)

If the NEX side is wrong, then I'll make the fix, as I've been working on that side

jonbarrow: The hash of the token is irrelevant, it's never seen by the client. The way the new token system works is tokens are completely random bytes, and those random bytes are what the client sees. The hash is stored in the database to avoid storing the token directly as a security measure

If you're asking if the token bytes themselves should be Nintendo base64 encoded when coming from NASC, yes that's the intent as it matches the official server implementation. But that's only for NASC, NNAS encodes the token differently (but NEX should already be detecting that?)

DaniElectra: What I mean is that the hash of the token here is calculated with the Nintendo base64. However, on the common protocols, there is an additional step before the hash which converts the Nintendo base64 to normal base64 before passing it to gRPC. This was inherited from the previous implementation, which would decode the token within common protocols.

Thus my question is whether the hash should be based on the Nintendo base64 or the normal base64 (I assume the former but asking just in case)

jonbarrow: Ah my bad. Yes, the former. It should be based on the N version not the converted version

DaniElectra (approved):
Tested the NEX side locally with no issues. LGTM

PR #317, opened 2026-01-11
Add basic health check before returning server address

jonbarrow:
Resolves #XXX

### Changes:

Adds a basic health check when requesting server information. Uses a simple race and whichever request returns first, wins. Falls back to a random IP if none return in time and just hopes for the best. I tested this with a local echo server and and a dedicated script and works as intended

### DIFF
--- a/src/models/server.ts
+++ b/src/models/server.ts
@@ -1,6 +1,49 @@
+import dgram from 'node:dgram';
+import crypto from 'node:crypto';
 import { Schema, model } from 'mongoose';
 import uniqueValidator from 'mongoose-unique-validator';
 import type { IServer, IServerConnectInfo, IServerMethods, ServerModel } from '@/types/mongoose/server';
+
+// * Kinda ugly to slap this in with the Mongoose stuff but it's fine for now
+// TODO - Maybe move this one day?
+const socket = dgram.createSocket('udp4');
+const pendingHealthCheckRequests = new Map<string, () => void>();
+
+socket.on('message', (msg: Buffer, _rinfo: dgram.RemoteInfo) => {
+	const uuid = msg.toString();
+	const resolve = pendingHealthCheckRequests.get(uuid);
+
+	if (resolve) {
+		resolve();
+	}
+});
+
+socket.bind();
+
+function healthCheck(target: { host: string; port: number }): Promise<string> {
+	return new Promise((resolve, reject) => {
+		const uuid = crypto.randomUUID();
+
+		const timeout = setTimeout(() => {
+			pendingHealthCheckRequests.delete(uuid);
+			reject(new Error('No valid response received'));
+		}, 5 * 1000); // TODO - Make this configurable? 5 seconds seems fine for now
+
+		pendingHealthCheckRequests.set(uuid, () => {
+			clearTimeout(timeout);
+			pendingHealthCheckRequests.delete(uuid);
+			resolve(target.host);
+		});
+
+		socket.send(Buffer.from(uuid), target.port, target.host, (error) => {
+			if (error) {
+				clearTimeout(timeout);
+				pendingHealthCheckRequests.delete(uuid);
+				reject(error);
+			}
+		});
+	});
+}
 
 const ServerSchema = new Schema<IServer, ServerModel, IServerMethods>({
 	client_id: String,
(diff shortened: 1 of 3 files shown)


src/models/server.ts
```diff
@@ -20,7 +63,8 @@ const ServerSchema = new Schema<IServer, ServerModel, IServerMethods>({
 	access_mode: String,
 	maintenance_mode: Boolean,
 	device: Number,
-	aes_key: String
+	aes_key: String,
+	health_check_port: Number
```
mrjvs: Let's make the new database field optional, we can't guarantee it exists for every server entry when this releases.

That also means the case of an undefined `health_check_port` needs to be handled. Perhaps just assume healthcheck succeeded if no port is set?

src/models/server.ts
```diff
@@ -31,9 +75,26 @@ ServerSchema.method('getServerConnectInfo', async function (): Promise<IServerCo
...
+
+	let target: string | undefined;
+
+	try {
+		// * Pick the first address that wins the health check. If no address responds in 5 seconds
+		// * nothing is returned
+		target = await Promise.race(healthCheckTargets.map(target => healthCheck(target)));
+	} catch {
+		// * Eat error for now, this means that no address responded in time
+		// TODO - Handle this
```
mrjvs: Let's add a log for it failing to find anything healthy, good for detecting failures for us sysadmin people

src/models/server.ts
```diff
@@ -31,9 +75,26 @@ ServerSchema.method('getServerConnectInfo', async function (): Promise<IServerCo
...
+		host: ip,
+		port: this.health_check_port
+	}));
+
+	let target: string | undefined;
+
+	try {
+		// * Pick the first address that wins the health check. If no address responds in 5 seconds
+		// * nothing is returned
+		target = await Promise.race(healthCheckTargets.map(target => healthCheck(target)));
```
mrjvs: I think we should take a random IP to test first, and then follow it up with a `Promise.race` after a failure of the first IP. This is to reduce the impact of downtime.

Take this example case:
- We have 5 ip addresses for `friends`
- We take a random IP to test first. meaning every IP has 20% of our users
- If one IP goes down, only 20% of users have to restart their console to get a good IP

If we only take the fastest IP:
- We have 5 ip addresses for `friends`
- We take a fastest IP, it will be the closest region IP, so it will very likely be 90% of users on the same IP
- The most common IP goes down, meaning 90% of users have to restart their console to get a good IP

src/models/server.ts
mrjvs: The `health_check_port` needs to be added to the provisioning config. Then the configuration of it will be fully automated by our infra.

mrjvs (changes requested):

src/models/server.ts
jonbarrow: I'm gonna be honest I have no idea what this provisioning config is. It was added without anyone tagging me or telling me anything, I didn't even see the PR

mrjvs: It loads a json file of partial server data and merges it into existing database entries. That way the infra can choose what ips and ports to use (the infra knows that info, so it's fully automated)

With the frequent IP downtimes we've had, I had to manually move a lot of servers, updating 43 db entries everytime. Now with the provisioning config, it's all automatic

src/models/server.ts
jonbarrow: Since all the IPs point to the same server, they should all be the "same" speed, no? When I tested this locally I saw a pretty even distribution of results. Doing the checks the way you described would slow down the endpoints up to possibly 10 seconds per request (5 seconds for the first check to fail, and then up to 5 seconds for the race), which is pretty significant when the initial connection is already somewhat slow. Using the race takes a small compromise in distribution in exchange for keeping the endpoints fast in the worst case scenario

mrjvs: Ips of the same machine can still be routed differently and will make some routes faster than others. In fact, we had to choose a datacenter to route it through when buying the IPs.

If you're worried about slowdowns when an IP goes down, I recommend:
- change the timeout to 2 seconds, a UDP pingpong shouldn't take longer than that
- do the race at the same time as the random check. But prefer the result from the random one if it succeeds.

With that combined, the worst case scenario is 2 seconds. And with the example of 5 IPs, the odds are 80%->fast, 20% -> 2s

jonbarrow: Ok

src/models/server.ts
jonbarrow: Works for me :+1:

src/models/server.ts
jonbarrow: Yes sir 🫡

mrjvs (approved):
Looks good, can't wait to apply to prod. Exciting stuff

PR #320, opened 2026-01-21
Add grace period for account deletions

### LINKED ISSUE #303: [Enhancement]: Add a grace period for the account deletion process to allow people to recover their deleted account
### Checked Existing

Currently, when you delete a PNID from the website, it will instantly delete the account, and it is very easy to do so (2 clicks!). The account deletion process should have a grace period of around 7 days so people can recover their PNID if they accidentally started the process or changed their mind. There have been various cases on the Pretendo Discord and the Forum where users accidentally clicked the Delete Account button and were unable to retrieve their account back, so giving users a grace period for account deletion allows users to recover their account if they choose to do so.

Examples:
https://canary.discord.com/channels/408718485913468928/1327488250226413618/1441777711667216534
https://canary.discord.com/channels/408718485913468928/1435624600183636101/1435624600183636101
(and countless others)

_No response_

jonbarrow:
Resolves #303

### Changes:

Adds new `marked_for_deletion` and `hard_delete_time` fields to the PNID model, and replaces all instances of `pnid.scrub()` with `pnid.markForDeletion()`. `pnid.markForDeletion()` sets `marked_for_deletion` to true and `hard_delete_time` to 7 days in the future. A new cron job runs every day at 2am (same as the BOSS server cron job) to run through all the pending deletions ready to be deleted, and *then* processes them. Also checks at server boot to catch any that happened during downtime, though once the cron job starts it shouldn't matter anyway

This lets us "rollback" any accidental deletions that may occur. There isn't a way for us to undo a pending deletion at the moment outside of direct DB access, but that should be fine for now since I don't think we'd be processing too many of those kinds of requests

Should be merged before https://github.com/PretendoNetwork/account/pull/313, since these changes can be deployed at any time


src/services/grpc/account/v2/delete-account.ts
```diff
@@ -17,7 +17,7 @@ export async function deleteAccount(request: DeleteAccountRequest): Promise<Dele
 	try {
 		const email = pnid.email.address;
 
-		await pnid.scrub();
+		pnid.markForDeletion();
```
DaniElectra, line 20: Shouldn't this to an `await`? This applies where this is called


```suggestion
		await pnid.markForDeletion();
```

DaniElectra (changes requested):
While we're here I think we should also update how the scrubbing works. Right now it scrubs the access level and sets it to 0, but this is kinda problematic especially for moderation, such as in the case where a PNID is banned and they delete their account.

I'd say to reset the level to 0 if it's above 0, but keep it as-is if it's lower than 0

src/services/grpc/account/v2/delete-account.ts
jonbarrow: Oops, that function isn't supposed to be async!

jonbarrow:
> While we're here I think we should also update how the scrubbing works. Right now it scrubs the access level and sets it to 0, but this is kinda problematic especially for moderation, such as in the case where a PNID is banned and they delete their account.
> 
> I'd say to reset the level to 0 if it's above 0, but keep it as-is if it's lower than 0

I forgot about this issue, sounds good to me

DaniElectra (approved):
LGTM

jonbarrow:
@mrjvs does this look good to you? Wanted to make sure the cron job and such won't mess with any deployments since it's the first time I've use a cron job since the new infra came out

mrjvs:
There's no problems with using nodejs cronjobs in the current infra setup.

If there is ever a need to scale the account server up to multiple instances, we may need a distributed locking system on top of cronjobs (so that we dont scrub the same account twice in the same millisecond)

jonbarrow:
> There's no problems with using nodejs cronjobs in the current infra setup.

Gotcha

> If there is ever a need to scale the account server up to multiple instances, we may need a distributed locking system on top of cronjobs (so that we dont scrub the same account twice in the same millisecond)

Is this a concern for now or do you think it can wait until later?

mrjvs:
It can wait, I don't think account currently can be horizontally scaled anyway. There's more work to be done before that's an option.

jonbarrow:
Sounds good then, thanks 👍

PR #326, opened 2026-04-19
Feat: Implement GetPNIDs rpc & stub newer account v2 grpc methods

CaramelKat:
Resolves #XXX

### Changes:

This implements the GetPNIDs rpc method that's needed for the Splatoon website. It also brings in the stubs for the new v2 account grpc api that was being implemented in the feat/grpc-tokens branch


jonbarrow:
My apologies, I did not see this PR until just now. Looks like Will beat you to some of the gRPC things which created some conflicts now

src/services/grpc/account/v2/get-pnids.ts
```diff
@@ -0,0 +1,119 @@
...
+import { PNID_PERMISSION_FLAGS } from '@/types/common/permission-flags';
+import { config } from '@/config-manager';
+import { Device } from '@/models/device';
+import type { GetPNIDsRequest, GetPNIDsResponse } from '@pretendonetwork/grpc/account/v2/get_pnids_rpc';
+
+export async function getPNIDs(request: GetPNIDsRequest): Promise<GetPNIDsResponse> {
+	const response: GetPNIDsResponse = { userData: [] };
+	for (const pid of request.pid) {
+		const pnid = await getPNIDByPID(pid);
+		if (!pnid) {
```
jonbarrow, line 11: This whole block should be impossible. This API is only usable by our systems, which would always know that a PID exists. So this whole part can just be skipped. Additionally there's no clear indication that an account is "unknown" besides some heuristics with this (like the username, but "Unknown" can be a valid account name, so the heuristics kinda fall apart)

It would be better to just omit unknown PIDs from the response entirely, rather than relying on heuristics like this. The response already contains the PID of the user in the data, so there's no real requirement for the output to be the same length as the input. If whatever system needs to know a PID is "invalid" then it can determine that on its end by checking if the output contains a key for input or not, which is an actual signal the caller can rely on and would also be simpler than having the caller heuristically check the objects

jonbarrow (changes requested):
Other than this, looks good to me. Though it is a bit redundant now to have all the methods implemented as stubbed, when Will's changes from dev that got pulled in already handle that in the implementation file. So those stubbed files in this PR can also be removed, though it's not an issue per-say

jonbarrow (dismissed):
LGTM, though idk what's up with those build errors?

CaramelKat:
Looks to be good now, apparently I missed pushing the lock file so fingers crossed this is the last review re-request on this >.>'

DaniElectra (approved):
LGTM!

PR #327, opened 2026-04-19
Unescape timezone list in NNAS

### LINKED ISSUE #325: [Bug]: Empty timezone field being returned during account creation results in 022-2932
### Checked Existing

### What happened?

Selecting certain time zones during the account creation process will result in the 3DS returning an empty tz_name field. This causes the account server to throw out an error and send a Bad Request [1600] to the 3DS. This shows up in the form of 022-2932, although it used to show up as 022-2402 before a pull request changed the error formatting.

I expect account creation to proceed as normal as it would on Nintendo Network.

### Steps to reproduce?

1. Attempt to create a new PNID through the Account Settings on a Nintendo 3DS
2. Select either "(UTC-05:00) Eastern Time (US & Canada)" or "(UTC-06:00) Central Time (US & Canada)" while filling out the time zone for your account.

The account server returns `[ERROR]: [POST] /v1/api/people: ValidatoinError: timezone.name: Cast to string failed for value "{}" (type Object) at path "timezone.name"` as a result of the empty time zone field.

This used to show up as 022-2402 until a pull request changed the formatting (potentially #306). This was the error code where we directed users to create their account while having their device language set to French. Attached below is a small portion of the request where the time zone was empty followed by the server response (this was on a local instance of Pretendo) to the request.

Image 

Image

jonbarrow:
Resolves #325

### Changes:

Re-applies https://github.com/PretendoNetwork/account/commit/4db3be92a0ec565f4363f0f106b7dc630517f807. These changes were reverted in https://github.com/PretendoNetwork/account/commit/1a3445db5beec02ae521dcb6dc0430bc437a06b1 and I forgot to bring them back. This should fix the issue with people getting 022-2932 on the 3DS when creating an account. See https://github.com/PretendoNetwork/account/issues/325 for details, but TLDR is that the JSON file was already escaping certain characters which caused the server to do a "double escape", and the 3DS didn't like this

### DIFF
--- a/src/services/nnas/timezones.json
+++ b/src/services/nnas/timezones.json
@@ -1983,7 +1983,7 @@
             {
                 "area": "America/Nassau",
                 "language": "en",
-                "name": "Eastern Time (US &amp; Canada)",
+                "name": "Eastern Time (US & Canada)",
                 "utc_offset": "-18000",
                 "order": "0"
             }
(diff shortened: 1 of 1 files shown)


DaniElectra (approved):
LGTM!

PR #334, opened 2026-05-08
Throw `0114` for users under 13

### LINKED ISSUE #333: [Bug]: Throw `0114` in NNAS when user is under 13
### Checked Existing

### What happened?

Currently NNAS does not check the users age on PNID registration, since as of Wii U firmware version 5.5.6 Nintendo update the Wii U to now display a message stating that a parent must call Nintendo directly to create a NNID for the user if the inputed age is under 13 and we were trusting that. However lower firmwares won't have this change

Also check on the server

### Steps to reproduce?

_No response_

See https://nintendo.wiki/wiki/Online/Nintendo_Network/NNAS#POST_/v1/api/people. I'm adding this, just creating the issue for tracking

jonbarrow:
Resolves #333 

### Changes:

Throws `0114` in NNAS (and equivalent in the API). This was missing because we trusted that the user would be on the latest firmware, in which Nintendo changed things with regard to COPPA and now require that users call Nintendo directly to create the account rather than using the old COPPA system. However older firmwares lack this, and we forgot to add it to the API in https://github.com/PretendoNetwork/account/pull/194 since we weren't accepting custom birthdays prior to that anyway, and custom/modded clients could bypass the client-side checks entirely anyway.

Rather than trying to implement the COPPA system, just block it all and call it a day. Too much of a headache

### DIFF
--- a/src/services/api/routes/v1/register.ts
+++ b/src/services/api/routes/v1/register.ts
@@ -118,6 +118,22 @@ router.post('/', async (request: express.Request, response: express.Response): P
 		}
 	}
 
+	if (age < 13) {
+		// * Wii U firmware 5.5.6 changed NNID setup to block setup of new accounts if the users age is
+		// * under 13, telling parents that they MUST call Nintendo to create the account, and we trusted
+		// * that users would be on these firmwares. Lower firmwares won't have this though, and will use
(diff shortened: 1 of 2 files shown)


DaniElectra (approved):
LGTM

PR #336, opened 2026-05-12
Fix automatic PNID scrubbing

jonbarrow:
Resolves #XXX

### Changes:

Removes the need to call `pnid.save()` during account deletions, this is now called internally. This also fixes a bug where the `checkMarkedDeletions` loop was not calling `pnid.save()` and thus automatic deletions were not being properly saved after the grace period was added (data like the forums and such was being deleted correctly, but the document data was not being saved).

Also adds a `try-catch` around the loop just to be safe.

### DIFF
--- a/src/database.ts
+++ b/src/database.ts
@@ -350,6 +350,10 @@ export async function checkMarkedDeletions(): Promise<void> {
 	});
 
 	for (const pnid of pnids) {
-		await pnid.scrub();
+		try {
+			await pnid.scrub();
+		} catch (error) {
+			LOG_ERROR(`Failed to scrub PNID ${pnid.pid}: ${error}`);
+		}
 	}
 }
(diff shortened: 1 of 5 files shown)


mrjvs (approved):

mrjvs:
Not for this PR, but at some the logger should probably get support for error logs. Atm it just stringifies it, which only gives the error message and not the stacktrace. So it makes it hard to see where the problem is coming from.

jonbarrow:
> Not for this PR, but at some the logger should probably get support for error logs. Atm it just stringifies it, which only gives the error message and not the stacktrace. So it makes it hard to see where the problem is coming from.

Agreed. At some point I was going to move logging to use Pino like Juxtaposition and such are using now. I just haven't gotten around to it because I figured it wasn't a super big priority, but I can make the move sooner if you'd like

The reason for the custom logger was because way back in the day there wasn't a solid Go logger that had the features we wanted, so we made plogger to do what we needed, and the idea was to implement the same kind of loggers for the other servers as well to keep them consistent, with the eventual plan being to make a logger parser for the admin panel

However this logger parser never happened, and this was all prior to us knowing about stuff like Pino, and now we use a totally different system for viewing logs, so it doesn't make sense to keep using the current logger

PR #346, opened 2026-07-26
fix: extend NEX token duration to 24 hours

### LINKED ISSUE #345: [Bug]: x01-0807 and x06-0807 resulting on game servers from expired NEX tokens
### Checked Existing

### What happened?

NEX Tokens are expiring too quickly for the average user, resulting in them being unable to log onto games and the friend list if they have had their console and/or game open for too long.

This seems to have popped up after the recent changes to how tokens are validated, and now seems to happen for every game and service. 

I expect there to be a failsafe for expired tokens where the console is made aware of it and can ask for a new. (Unsure if this is even possible)

### Steps to reproduce?

1. Log onto the friends list or a game
2. Wait and then disconnect (whether it's forced via the 3DS' wireless switch for friends list or normally disconnecting for a game)
3. Attempt to reconnect to the service

I'm unsure if this issue would be better for the account repository (where tokens are handled) or if this should go on nex-protocols-common-go. If it is on the wrong repository, feel free to move it.

This can also result in 015-5016 in the event it is a service token for Juxt.

BDMCGaming:
Resolves #345 

### Changes:

This is intended to be a temporary fix for NEX tokens expiring until a more concrete solution can be made. This should resolve x01-0807, x06-0807, and the occasional 015-5016 that all usually result from a NEX token being expired. Changes were tested through verifying that expiration date is longer via MongoDB Compass

- [X] I have read and agreed to the [Code of Conduct](https://github.com/PretendoNetwork/Pretendo/blob/master/.github/CODE_OF_CONDUCT.md).
- [X] I have read and complied with the [contributing guidelines](https://github.com/PretendoNetwork/Pretendo/blob/master/.github/CONTRIBUTING.md).
- [X] What I'm implementing was an [approved issue](../issues?q=is%3Aopen+is%3Aissue+label%3Aapproved).
- [X] I have tested all of my changes.

Notes:

We should look into how Nintendo handled token expirations. They definitely had some way of handling it, as I could keep my 3DS on for weeks at a time without having to reboot it due to an expired token. Did they really just not let NEX tokens expire, or did they have some sort of way to make the console regenerate?

### DIFF
--- a/src/services/nasc/routes/ac.ts
+++ b/src/services/nasc/routes/ac.ts
@@ -88,7 +88,7 @@ async function processLoginRequest(server: HydratedServerDocument, pid: number, 
 			token_type: TokenType.NEX,
 			title_id: BigInt(parseInt(titleID, 16)),
 			issued: new Date(),
-			expires: new Date(Date.now() + (3600 * 1000))
+			expires: new Date(Date.now() + (24 * 3600 * 1000))
 		}
 	});
 
--- a/src/services/nnas/routes/provider.ts
+++ b/src/services/nnas/routes/provider.ts
@@ -234,7 +234,7 @@ router.get('/nex_token/@me', async (request: express.Request, response: express.
 			token_type: TokenType.NEX,
 			title_id: BigInt(parseInt(titleID, 16)),
 			issued: new Date(),
-			expires: new Date(Date.now() + (3600 * 1000))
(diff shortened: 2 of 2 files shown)


mrjvs:
A token expiry PR has been merged on the NEX server end: https://github.com/PretendoNetwork/nex-protocols-common-go/pull/68

PR #351, opened 2026-08-19
Feat: account edit grpc endpoint

limesdotpink:
### Changes:

- add updateUserData method to grpc v2;
- remove incorrect types from PNID

- [X] I have read and agreed to the [Code of Conduct](https://github.com/PretendoNetwork/Pretendo/blob/master/.github/CODE_OF_CONDUCT.md).
- [X] I have read and complied with the [contributing guidelines](https://github.com/PretendoNetwork/Pretendo/blob/master/.github/CONTRIBUTING.md).
- [X] I have tested all of my changes.

### DIFF
--- a/src/services/grpc/api/v2/update-user-data.ts
+++ b/src/services/grpc/api/v2/update-user-data.ts
@@ -1,14 +1,155 @@
+import { ServerError, Status } from 'nice-grpc';
+import Mii from 'mii-js';
+import { isValidBirthday } from '@/util';
 import { config } from '@/config-manager';
+import timezones from '@/services/nnas/timezones.json';
+import regionsList from '@/services/nnas/regions.json';
 import type { CallContext } from 'nice-grpc';
-import type { UpdateUserDataRequest, DeepPartial } from '@pretendonetwork/grpc/api/v2/update_user_data_rpc';
+import type {
+	UpdateUserDataRequest,
+	DeepPartial
+} from '@pretendonetwork/grpc/api/v2/update_user_data_rpc';
 import type { GetUserDataResponse } from '@pretendonetwork/grpc/api/v2/get_user_data_rpc';
 import type { AuthenticationCallContextExt } from '@/services/grpc/api/v1/authentication-middleware';
 
-export async function updateUserData(_request: UpdateUserDataRequest, context: CallContext & AuthenticationCallContextExt): Promise<DeepPartial<GetUserDataResponse>> {
+export async function updateUserData(
+	request: UpdateUserDataRequest,
+	context: CallContext & AuthenticationCallContextExt
+): Promise<DeepPartial<GetUserDataResponse>> {
 	// * This is asserted in authentication-middleware, we know this is never null
 	const pnid = context.pnid!;
 
-	// TODO - STUBBED, DO SOMETHING HERE
+	const serverAccessLevel = request.serverAccessLevel?.trim();
+	const mii = request?.mii?.trim();
+	const birthday = request.birthday?.trim();
+	const gender = request.gender?.trim();
+	const country = request.country?.trim();
+	const region = request.region;
+	const timezone = request.timezone?.trim();
+
+	/* TODO: implement these if/when needed */
+	// const language = request.language?.trim();
+	// const marketingFlag = request.marketingFlag;
+
+	if (serverAccessLevel) {
+		if (['prod', 'test', 'dev'].includes(serverAccessLevel)) {
+			if (pnid.access_level > 0 && pnid.access_level < 4) {
+				if (
+					(serverAccessLevel === 'test' && pnid.access_level < 1) ||
+					(serverAccessLevel === 'dev' && pnid.access_level < 3)
+				) {
+					throw new ServerError(
+						Status.INVALID_ARGUMENT,
+						`Do not have permission to enter this environment`
+					);
+				}
+
+				pnid.server_access_level = serverAccessLevel;
+			} else {
+				throw new ServerError(Status.PERMISSION_DENIED, `Banned.`);
+			}
+		} else {
+			throw new ServerError(
+				Status.INVALID_ARGUMENT,
+				`Must be one of: prod, test, dev`
+			);
+		}
+	}
+
+	if (birthday) {
+		if (isValidBirthday(birthday)) {
+			pnid.birthdate = birthday;
+		} else {
+			throw new ServerError(
+				Status.INVALID_ARGUMENT,
+				`Must be a valid date formatted as: YYYY-MM-DD`
+			);
+		}
+	}
+
+	if (gender) {
+		if (['M', 'F'].includes(gender)) {
+			pnid.gender = gender;
+		} else {
+			throw new ServerError(
+				Status.INVALID_ARGUMENT,
+				`Must be one of: F, M`
+			);
+		}
+	}
+
+	if (country || region) {
+		let countryId = 0;
+
+		// if we have a region but no country, we extract the country id from it
+		if (region && !country) {
+			const regionHex = region.toString(16).padStart(8, '0');
+			countryId = parseInt(regionHex.slice(0, 2), 16);
+		}
+
+		const countryObj = regionsList.find((c) => {
+			return c.iso_code === country || c.id === countryId;
+		});
+
+		if (!countryObj?.iso_code) {
+			throw new ServerError(Status.INVALID_ARGUMENT, `Invalid country`);
+		}
+
+		if (region) {
+			const regionObj = countryObj.regions.find((r) => {
+				return r.id === region;
+			});
+
+			if (!regionObj) {
+				throw new ServerError(
+					Status.INVALID_ARGUMENT,
+					`Invalid region`
+				);
+			} else {
+				pnid.country = countryObj.iso_code;
+				pnid.region = region;
+			}
+		} else if (pnid.country !== countryObj?.iso_code) {
+			// if editing the country with no explicit region, set it to Unspecified
+			pnid.country = countryObj.iso_code;
+			pnid.region = countryObj.regions.find(
+				r => r.name === 'Unspecified'
+			)!.id;
+		}
+	}
+
+	if (timezone) {
+		const pnidCountryTimezones =
+			timezones[pnid.country as keyof typeof timezones];
+		// using japanese because some timezones are only available in that locale
+		const newTimezone = pnidCountryTimezones.ja.find(
+			t => t.area === timezone
+		);
+		if (newTimezone) {
+			pnid.timezone.name = newTimezone.area;
+			pnid.timezone.offset = Number(newTimezone.utc_offset);
+		} else {
+			throw new ServerError(Status.INVALID_ARGUMENT, `Invalid timezone`);
+		}
+	}
+
+	if (mii) {
+		try {
+			const parsedMii: Mii = new Mii(Buffer.from(mii, 'base64'));
+
+			parsedMii.validate();
+
+			await pnid.updateMii({
+				name: parsedMii.miiName,
+				primary: 'Y',
+				data: parsedMii.encode().toString('base64')
+			});
+		} catch {
+			throw new ServerError(Status.INVALID_ARGUMENT, `Invalid mii data`);
+		}
+	}
+
+	await pnid.save();
 
 	return {
 		deleted: pnid.deleted || pnid.marked_for_deletion,
@@ -26,6 +167,7 @@ export async function updateUserData(_request: UpdateUserDataRequest, context: C
 		birthday: pnid.birthdate,
 		gender: pnid.gender,
 		country: pnid.country,
+		region: pnid.region,
 		timezone: pnid.timezone.name,
 		language: pnid.language,
 		emailAddress: pnid.email.address,
@@ -39,7 +181,9 @@ export async function updateUserData(_request: UpdateUserDataRequest, context: C
 				priceId: pnid.connections.stripe.price_id,
 				tierLevel: pnid.connections.stripe.tier_level,
 				tierName: pnid.connections.stripe.tier_name,
-				latestWebhookTimestamp: BigInt(pnid.connections.stripe.latest_webhook_timestamp ?? 0)
+				latestWebhookTimestamp: BigInt(
+					pnid.connections.stripe.latest_webhook_timestamp ?? 0
+				)
 			}
 		},
 		marketingFlag: pnid.flags.marketing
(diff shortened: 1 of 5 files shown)


limesdotpink:
jvs be nice

src/services/grpc/api/v2/update-user-data.ts
```diff
@@ -1,14 +1,155 @@
...
+					(serverAccessLevel === 'test' && pnid.access_level < 1) ||
+					(serverAccessLevel === 'dev' && pnid.access_level < 3)
+				) {
+					throw new ServerError(
+						Status.INVALID_ARGUMENT,
+						`Do not have permission to enter this environment`
+					);
+				}
+
+				pnid.server_access_level = serverAccessLevel;
```
mrjvs: I'd rather have the specific levels checked rather than a forbidden check at the start.

Like this:
```ts
if (level==test) {
  if (wrongAccessLevel) throw;
  update
}

if (level==test) {
  if (wrongAccessLevel) throw;
  update
}

if (level==test) {
  if (wrongAccessLevel) throw;
  update
}
```

That makes it very explicit what conditions are checked per level and wont make it break when a new value gets introduced.

src/services/grpc/api/v2/update-user-data.ts
```diff
@@ -1,14 +1,155 @@
...
+	if (gender) {
+		if (['M', 'F'].includes(gender)) {
+			pnid.gender = gender;
+		} else {
+			throw new ServerError(
+				Status.INVALID_ARGUMENT,
+				`Must be one of: F, M`
+			);
+		}
+	}
```
mrjvs, line 92: I think I'd prefer guard clauses instead of conditionals: https://refactoring.guru/replace-nested-conditional-with-guard-clauses

src/services/grpc/api/v2/update-user-data.ts
```diff
@@ -1,14 +1,155 @@
...
+			pnid.timezone.name = newTimezone.area;
+			pnid.timezone.offset = Number(newTimezone.utc_offset);
+		} else {
+			throw new ServerError(Status.INVALID_ARGUMENT, `Invalid timezone`);
+		}
+	}
+
+	if (mii) {
+		try {
+			const parsedMii: Mii = new Mii(Buffer.from(mii, 'base64'));
```
mrjvs: Don't need a type indication here. it can be inferred.

src/services/grpc/api/v2/update-user-data.ts
mrjvs: (with most of these cases tbh)

src/services/grpc/api/v2/update-user-data.ts
```diff
@@ -1,14 +1,155 @@
...
+			} else {
+				pnid.country = countryObj.iso_code;
+				pnid.region = region;
+			}
+		} else if (pnid.country !== countryObj?.iso_code) {
+			// if editing the country with no explicit region, set it to Unspecified
+			pnid.country = countryObj.iso_code;
+			pnid.region = countryObj.regions.find(
+				r => r.name === 'Unspecified'
+			)!.id;
```
mrjvs: I'd rather have a throw rather than a non-null assertions

mrjvs (dismissed):

mrjvs (commented):
Lgtm, but still needs a stamp of approval from someone who knows how accounts work (👀 @jonbarrow)

limesdotpink (commented):
LGBT 👍

src/services/grpc/api/v2/update-user-data.ts
```diff
@@ -1,14 +1,176 @@
...
+	}
+
+	if (country || region) {
+		let countryId = 0;
+
+		// if we have a region but no country, we extract the country id from it
+		if (region && !country) {
+			const regionHex = region.toString(16).padStart(8, '0');
+			countryId = parseInt(regionHex.slice(0, 2), 16);
+		}
```
jonbarrow: In what cases is this happening? There shouldn't really be a case where a country is not present with its region. I haven't seen the changes on the website side of things yet, is this a result of that?

Either way, no need to allocate and manipulate strings. `(region >>> 24) & 0xFF` will get the first byte of the number

src/services/grpc/api/v2/update-user-data.ts
```diff
@@ -1,14 +1,176 @@
...
+			throw new ServerError(
+				Status.INVALID_ARGUMENT,
+				`Must be one of: F, M`
+			);
+		}
+
+		pnid.gender = gender;
+	}
+
+	if (country || region) {
```
jonbarrow: My 2 cents here: this whole section feels a bit unnecessary. When you break it down, this whole chunk just sets the users country ISO code and region value as seen in https://nintendo.wiki/wiki/Online/Nintendo_Network/Region_IDs

But if the website just sends the raw region value anyway, the only thing this section as to do to get the country ISO code and verify the region is valid is loop over the regions list once, skipping the need to extract the country ID from the region and all these other checks. Throw in a double loop and a `break label` and this is all it should need:

```js
const regions = require('./regions.json');

const targetRegion = 0xBA010000; // * Bermuda, highest possible region value and the worst case scenario
let foundCountry;
let foundRegion;

countryLoop: for (const country of regions) {
	for (const region of country.regions) {
		if (region.id === targetRegion) {
			foundCountry = country;
			foundRegion = region;

			break countryLoop;
		}
	}
}

console.log(foundCountry, foundRegion);
```

When I timed this it only took \~0.3ms to loop over the entire set of regions for the worst case scenario and should have the same perf as the current version in terms of loop speed, and it gets you the same data as before

src/services/grpc/api/v2/update-user-data.ts
```diff
@@ -1,14 +1,176 @@
...
+			t => t.area === timezone
+		);
+
+		if (!newTimezone) {
+			throw new ServerError(Status.INVALID_ARGUMENT, `Invalid timezone`);
+		}
+
+		pnid.timezone.name = newTimezone.area;
+		pnid.timezone.offset = Number(newTimezone.utc_offset);
+	}
```
jonbarrow, line 123: This feels like a bug we need to address at some point, can we note that? The "only using Japan" part feels smelly

jonbarrow (commented):
Other than this, did you run the linter? I thought the linter was setup to complain about stuff like using template literals without expressions

I'll do a linter/formatting update later, not in this PR, just asking to see if the linter is working correctly

src/services/grpc/api/v2/update-user-data.ts
limesdotpink: My implementation also supports setting a country code via iso code (e.g. `US`, `NL`, ...). If that's something we don't care for, it can be removed (at that point we don't even need to have the country in the proto, so that can be removed as well).

jonbarrow: I think basing it off the region ID entirely is probably the easiest way to go tbh. It simplifies the checks on the account server and the same region list can be chucked into the website to power that side too

src/services/grpc/api/v2/update-user-data.ts
limesdotpink: It's not happening, but since it's a two-liner (and now a one-liner), I think it's fine to keep it, since it *is* a valid case. Like I said in the reply to your other comment, the api accepts both a country code (e.g. `US`, `IT`) and a region id (as present in the `regions.json`). If an explicit country code is not provided, the country id is extracted from the region field.

If we choose to keep this as is, I'll definitely go with your core concise way to extract that.

src/services/grpc/api/v2/update-user-data.ts
limesdotpink: Sure. Want me to make an issue in the [wiiu-country-region-extractor repo](https://github.com/PretendoNetwork/wiiu-country-region-extractor) for it?

An example of a region with japanese only is Fiji:
```json
"FJ": {
    "ja": [
        {
            "area": "Pacific/Fiji",
            "language": "ja",
            "name": "フィジー、マーシャル諸島",
            "utc_offset": "43200",
            "order": "0"
        }
    ]
}
```

src/services/grpc/api/v2/update-user-data.ts
limesdotpink: Fair enough!

limesdotpink:
> Other than this, did you run the linter? I thought the linter was setup to complain about stuff like using template literals without expressions

cc @jonbarrow I did :/

src/services/grpc/api/v2/update-user-data.ts
jonbarrow: That repo is only used for extracting the region data for the region IDs because there isn't really any other consumable source for that data, and we needed it

The timezones though are something I scraped from the official NNAS server managed by Nintendo. I had it run through every 2-letter combination, twice (once for the country once for the language), and download the contents of https://account.nintendo.net/v1/api/content/time_zones/COUNTRY/LANGUAGE

So either something bugged out in my scraping, or something else is going on

We could probably just do the same thing here as the region/country section though. Throw this same list of timezones into the website, so that the website has the same list as NNAS, and then just make sure the value from the client matches one NNAS knows about

jonbarrow:
> > Other than this, did you run the linter? I thought the linter was setup to complain about stuff like using template literals without expressions
> 
> cc @jonbarrow I did :/

All good, not an issue with you or anything. Was just making sure, so I know what to do when I make a "linter update" PR 👍

src/types/mongoose/pnid.ts
```diff
@@ -32,8 +32,6 @@ export interface IPNID {
 	timezone: {
 		name: string;
 		offset: number;
-		marketing: boolean;
-		off_device: boolean;
```
jonbarrow, line 36: Sorry for not mentioning this in my last review, just noticed this. Why did you remove these? We do still want to use these in the future, and you only removed them from the types but not the models?

The `marketing` flag is what determines if the user can get unsolicited emails (Nintendo used this for marketing, hence the name, like eShop emails and stuff, but we had planned to use it for like newsletters and emails about new blog posts and stuff)

The `off_device` flag is what determines if the account can be logged into outside of the Wii U/3DS (PC, web, mobile, etc.), the OAuth "login with NNID" stuff, which is going to be needed for 3rd party service integrations

src/services/grpc/api/v2/update-user-data.ts
binaryoverload: The "area" seems to just TZ timezones?

https://en.wikipedia.org/wiki/List_of_tz_database_time_zones

We could use a standard library to parse them: https://github.com/date-fns/date-fns/tree/main/pkgs/tz since we seem to only use the name and offset

src/types/mongoose/pnid.ts
mrjvs: It was removed from the `timezone` field. The correct place at `flags` still exist.

The model never had the fields under `timezone`, it was likely an early copy paste error when porting to typescript.

jonbarrow: Ah okay, thank you. My apologies. The diff made it look like it was just removed entirely, I didn't check the surrounding lines

src/services/grpc/api/v2/update-user-data.ts
jonbarrow: > The "area" seems to just TZ timezones?
> 
> [en.wikipedia.org/wiki/List_of_tz_database_time_zones](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones)
> 
> We could use a standard library to parse them: [date-fns/date-fns@`main`/pkgs/tz](https://github.com/date-fns/date-fns/tree/main/pkgs/tz?rgh-link-date=2026-08-20T14%3A03%3A09Z) since we seem to only use the name and offset

I've tried a few times in the past to use a standard library for timezone handling, but could never find one that 1-to-1 matched Nintendo's. Which is why we've just always used this list instead

The consoles are very picky about things, so for that reason (and for the sake of being accurate) I try to match Nintendo's output 1-to-1 where possible. The timezone stuff can even [crash the console](https://github.com/PretendoNetwork/account/blob/8c802d7f6c4825179f1d42533e08893f7a80bcd6/src/services/nnas/routes/content.ts#L134-L150) if not done correctly

limesdotpink: re: https://github.com/PretendoNetwork/account/pull/351#discussion_r3822157330
Oops, I had just assumed that the timezone json came from that too, my bad.

> We could probably just do the same thing here as the region/country section though. Throw this same list of timezones into the website, so that the website has the same list as NNAS, and then just make sure the value from the client matches one NNAS knows about

yeah, that's pretty much what I'm doing already

re: https://github.com/PretendoNetwork/account/pull/351#discussion_r3822252417

That's definitely an option for *parsing* timezones, but we're gonna need the timezones around for NNAS (since the account-settings page needs to display a list of timezones for the user to pick from), so we might as well use it?

binaryoverload: Makes sense 

Sounds like the timezones file might just need some clean up to ensure we have at least the English translation of all timezones, and then use that as the base - similar to how the translation system works

binaryoverload:
> All good, not an issue with you or anything. Was just making sure, so I know what to do when I make a "linter update" PR 👍

@jonbarrow I've made a common-configs PR to address this: https://github.com/PretendoNetwork/common-configs/pull/7

src/services/grpc/api/v2/update-user-data.ts
jonbarrow: > Makes sense
> 
> Sounds like the timezones file might just need some clean up to ensure we have at least the English translation of all timezones, and then use that as the base - similar to how the translation system works

Doing that might have other issues, since it's not what the console expects. I did just check the official servers and yeah for some reason I can only get it to return Japanese for Fiji, I guess that's just How It Is:tm: for some reason

That being said, this issue would be alleviated by using the same system as now used by the country/regions. Limit the website to being able to use the same list as served by NNAS, and then just making sure the user input is one that exists in this list. That both simplifies the checks, and removes the hacky work around, and doesn't require us to change the timezone list in a way that doesn't match Nintendo's output

binaryoverload: I think we're referring to two separate issues:
- A timezone not existing in the NNAS list (Fixed by making the website respect the NNAS list)
- A timezone that only exists in one locale (hence the hack)

At the moment, this PR can only aim to fix the hack for the one locale issue. 

It would be slightly slower, but a less hacky way would be to make the code iterate over the locales and get the first that matches (Since we're only looking for the numeric value here, which will be the same in all locales)

limesdotpink:
@jonbarrow think we can get this merged?

jonbarrow (approved):

jonbarrow:
> @jonbarrow think we can get this merged?

No

PR #352, opened 2026-08-23
Add sanity checks to `/v1/api/people` NNAS endpoint

jonbarrow:
Resolves #XXX

### Changes:

Adds sanity checks to the `/v1/api/people` endpoint on the NNAS service. This includes porting the various username/password checks from the gRPC/API services to this endpoint, as well as checking the other input data

The error codes/responses match the current documentation found at https://nintendo.wiki/wiki/Online/Nintendo_Network/NNAS#POST_/v1/api/people

Also centralizes the PNID username/password rule checks into util, so that we don't have to keep duplicating them between 3 services. I used the term "NNID" in that context because its the rules Nintendo set, not something we did, so it felt weird using the "PNID" name


src/services/nnas/routes/people.ts
```diff
@@ -108,12 +151,84 @@ router.post('/', deviceRatelimit, deviceCertificateMiddleware, async (request: e
...
+	if (!person.user_id) {
+		response.status(400).send(xmlbuilder.create({
+			errors: {
+				error: {
+					cause: 'userId',
+					code: '1104',
+					message: 'User ID format is not valid'
+				}
+			}
+		}).end());
```
binaryoverload: I think there should be a function that will build these sort of errors (including 0002) - since 9 lines repeated for all the checks is quite a lot

binaryoverload (changes requested):

src/services/nnas/routes/people.ts
binaryoverload: From the looks of the structure, it could be something like:

```js
//         code  message                        cause (opt)
buildError(1104, "User ID format is not valid", 'userId')
```

jonbarrow: I have a branch/PR with a total overhaul to this system already, it just has merge conflicts and needs to be updated https://github.com/PretendoNetwork/account/pull/170. I'd prefer to keep that work there rather than trying to mix different implementations, and keep this PR scoped to the sanity checks themselves

binaryoverload: Looking at that PR, IMO it doesn't actually improve it a whole lot (apart from type safety)

For the majority of errors, we can do a 1 liner as above

jonbarrow: The details of that can be discussed on that PR, but like I said I'd like to keep this PR scoped to the sanity checks themselves so I don't have to keep overwriting things in the other branch once I pick it back up. It has enough merge conflicts as it is, and the one liner you wrote doesn't take into account situations like when the endpoint needs to return multiple errors and such

binaryoverload: Sure 👍

src/services/nnas/routes/people.ts
```diff
@@ -3,19 +3,32 @@ import express from 'express';
...
+const PNID_VALID_CHARACTERS_REGEX = /^[\w\-.]*$/;
+const PNID_PUNCTUATION_START_REGEX = /^[_\-.]/;
+const PNID_PUNCTUATION_END_REGEX = /[_\-.]$/;
+const PNID_PUNCTUATION_DUPLICATE_REGEX = /[_\-.]{2,}/;
+
+// * This sucks
+const PASSWORD_WORD_OR_NUMBER_REGEX = /(?=.*[a-zA-Z])(?=.*\d).*/;
+const PASSWORD_WORD_OR_PUNCTUATION_REGEX = /(?=.*[a-zA-Z])(?=.*[_\-.]).*/;
+const PASSWORD_NUMBER_OR_PUNCTUATION_REGEX = /(?=.*\d)(?=.*[_\-.]).*/;
+const PASSWORD_REPEATED_CHARACTER_REGEX = /(.)\1\1/;
```
binaryoverload: This seems to be the same copy-pasted thing from the other places? Could we make this one  set of constants?

src/services/nnas/routes/people.ts
```diff
@@ -123,6 +238,305 @@ router.post('/', deviceRatelimit, deviceCertificateMiddleware, async (request: e
...
+					message: 'Password cannot be the same as User ID'
+				}
+			}
+		}).end());
+
+		return;
+	}
+
+	// TODO - Check against the list of valid countries
+	if (!person.country) {
```
binaryoverload: Is the country a country code or a name? If it's a country code, that could be validated to be at least the right chars

src/services/nnas/routes/people.ts
```diff
@@ -123,6 +238,305 @@ router.post('/', deviceRatelimit, deviceCertificateMiddleware, async (request: e
...
+	// * 				message: 'Account country and device country do not match'
+	// * 			}
+	// * 		}
+	// * 	}).end());
+	// *
+	// * 	return;
+	// * }
+
+	// TODO - Check against the list of valid languages
+	if (!person.language) {
```
binaryoverload: Same comment as the country code

binaryoverload (changes requested):

src/services/nnas/routes/people.ts
jonbarrow: I ***THINK*** it's just the 2 letter ISO country code. But I have not confirmed this and we don't have a full list of known valid countries (it's the same list that applies to the `X-Nintendo-Country` header though)

You can see the full request body structure here https://nintendo.wiki/wiki/Online/Nintendo_Network/NNAS#POST_/v1/api/people

Edit: actually looks like Cemu has a full list of valid countries, I'll dig into this a bit more

src/services/nnas/routes/people.ts
jonbarrow: Same reply as the country code /s

src/services/nnas/routes/people.ts
jonbarrow: I was also thinking about this, but I genuinely don't where the best place to put it would be since we don't really have a standard place to put constants used by multiple services

binaryoverload: Probably just https://github.com/PretendoNetwork/account/blob/dev/src/util.ts ?

src/services/nnas/routes/people.ts
jonbarrow: Okay I think I got it, I believe this is just the allowed languages from the timezone list, which also matches up with the standard set of languages Nintendo uses elsewhere. I'll implement it under that assumption and if stuff breaks later we can patch it

src/services/nnas/routes/people.ts
binaryoverload: I was thinking that if we know the format (which if it's a code should always be two chars) then we can validate the format without the list of countries

just `[A-Z]{2}`

src/services/nnas/routes/people.ts
jonbarrow: Works for me

jonbarrow: I went ahead and just centralized everything into some helpers inside util, rather than exporting the constants and duplicating the rule checks. It adds quite a few functions but it cleans up the call sites quite a bit imo

src/services/nnas/routes/people.ts
```diff
@@ -3,19 +3,181 @@ import express from 'express';
...
+	'fr',
+	'de',
+	'it',
+	'es',
+	// * Likely simplified Chinese, but we haven't seen this in practice
+	// * Likely Korean, but we haven't seen this in practice
+	'nl',
+	'pt',
+	'ru'
+	// * Likely traditional Chinese, but we haven't seen this in practice
```
DaniElectra, line 178: The languages missing didn't have NN support

jonbarrow: Ye, that's what I meant when I said we haven't seen it

jonbarrow:
I believe I solved the merge conflicts, is this good to go now?

jonbarrow:
Once this is merged I should be good to go on finishing the audit log stuff, I was going to start earlier today but realized this PR touches some areas of that and I'd rather not create more merge conflicts

DaniElectra (approved):
lgbt

jonbarrow:
> lgbt

Sweet! @binaryoverload can I get you to look over this too, since the merge is blocked from your last review?

binaryoverload (approved):

PR #358, opened 2026-09-01
Healthchecks, metrics & more

mrjvs:
Changes:
- Added prometheus metrics integration
  - On a separate express server for security
  - Not enabled by default, must be configured for it to work
- Added some metrics:
  - Default express metrics, so we get to see a bunch of HTTP metrics by default
  - PNID total
  - NEX Token total
- Removed account deletion workflow from boot, only does it on the schedule now. This makes account server boot faster and not cause downtime.
- Added a healthcheck endpoint, so if the server takes forever to boot it doesn't cause downtime and simply waits for it to be done before sending it traffic.

And a bunch of small stuff:
- Removed used of deprecated field `req.host`
- Made schedule creation not an error but a success log
- Made exceptions on HTTP endpoints log the full error, not just the message

### DIFF
--- a/src/metrics.ts
+++ b/src/metrics.ts
@@ -1,0 +1,47 @@
+import { format } from 'node:util';
+import expressMetrics from 'express-prom-bundle';
+import express from 'express';
+import { LOG_ERROR, LOG_INFO, LOG_SUCCESS } from '@/logger';
+import { config } from '@/config-manager';
+import type { Express, NextFunction, Request, Response } from 'express';
+
+export function registerMetrics(app: Express): Express {
+	const metrics = express();
+
+	if (config.metrics.enabled) {
+		LOG_INFO('Setting up metrics');
+		app.use(expressMetrics({
+			// * Include full express and nodejs metrics
+			includeMethod: true,
+			includePath: true,
+			urlValueParser: {
+				minBase64Length: 15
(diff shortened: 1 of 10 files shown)


src/metrics.ts
```diff
@@ -0,0 +1,75 @@
...
+import { config } from '@/config-manager';
+import { PNID } from '@/models/pnid';
+import { NEXToken } from '@/models/nex-token';
+import type { Express, NextFunction, Request, Response } from 'express';
+
+export const pnidTotalGauge = new Gauge({
+	name: 'pn_account_pnid_total',
+	help: 'Total number of registered PNIDs',
+	async collect(): Promise<void> {
+		// * Aggregations are faster on large collections
```
jonbarrow, line 15: Is this actually faster than `estimatedDocumentCount`?

jonbarrow (approved):
Just one question but other than that, LGBT 🫡

src/metrics.ts
mrjvs: It's faster than `.countDocument({})`, it solved a lot of the timeouts we had in juxt.

`estimatedDocumentCount()` will be faster cuz its just a cached number in mongo, but it doesn't support filtering (like getting the deleted pnids out of it)

PR #359, opened 2026-09-16
Unstub audit log gRPC methods

jonbarrow:
Resolves #XXX

### Changes:

Unstubs the relevant gRPC methods relating to audit logs.

The scope of this PR is intentionally limited to ***just*** the audit log sections. The goal is not to completely unstub the entire service, there are more methods (like listing bans and such) which are still left stubbed as they do not directly contribute to audit logs. The only minor exception to this is bans, where I also updated NASC/NNAS to check for the new ban system

This also relies on the modern permission system, so anyone who needs access to these methods will need to be given the new permissions in the admin panel

Older APIs have also been updated to detect the presence of bans from the modern ban API implemented here and will seamlessly map a users legacy access level based on the bans, so older clients should be able to function with the new ban system just fine


src/services/grpc/account/v2/issue-ban.ts
```diff
@@ -1,9 +1,196 @@
...
+};
+
+const SCOPE_TYPES = [
+	BanScopeType.BAN_SCOPE_TYPE_ALL,
+	BanScopeType.BAN_SCOPE_TYPE_APPLICATION,
+	BanScopeType.BAN_SCOPE_TYPE_NEX_SERVICE,
+	BanScopeType.BAN_SCOPE_TYPE_INDEPENDENT_SERVICE
+];
+
+export async function issueBan(request: IssueBanRequest, context: CallContext): Promise<IssueBanResponse> {
```
mrjvs, line 42: If bans are being implemented, we also needs some kind of way for dependent services to view that data.

I think there are three places where bans are currently being used:
- Website shows a banned label on a users profile
- Juxt shows network ban state on the moderation profile page
- Users are prevented from using juxt if they are network banned

mrjvs (changes requested):

src/services/grpc/account/v2/issue-ban.ts
jonbarrow: This can be done with the [`ListBans` RPC](https://github.com/PretendoNetwork/grpc/blob/10909a868ea7f85fd39d56c28d3a07f2e1d452c5/protobufs/account/v2/list_bans_rpc.proto), as it takes in filters to list bans by a specific target. Since that doesn't have any bearing on audit logs, I intentionally did not implement that RPC in this PR to keep the scope limited. This new ban system hasn't replaced the old one, so bans will continue to function as normal until the admin panel is changed to specifically call `IssueBan`

mrjvs: Sounds good, we can delay the implementation until then.

However ListBans will be quite inefficient. Juxt would have to call this on every request. I think to prevent breaking changes, forcing the `access_level` field to be -1 if a network ban is active will be beneficial.

jonbarrow: Juxt is already doing that to get the user data, I don't think it would be that much of an impact to add one extra request especially since gRPC is quite efficient. I can add that suggestion, but I consider that a hack personally, ideally all services should migrate to the modern APIs eventually

DaniElectra: I agree with jvs on changing the access level on API responses for compatibility. As much as we'd like to get every service to use the new API, there are some where that might not be possible without big changes (see MK8)

jonbarrow: Like I said I can add that suggestion, I just consider it a hack that should ideally be migrated away from in other services rather than something that keeps getting relied on in future updates/services

jonbarrow: @mrjvs @DaniElectra I've added this in what I believe the cleanest way would be. Rather than having the new ban API modify the legacy value, I updated the older APIs to also look for modern bans and override the access level in the response with the older `-1` value if any modern bans exist

The issue with having the newer ban system physically change user/device access levels like that is that the older system has no concept of things like temp bans, and it means that any newer tools which pardon modern bans would also need to restore the old users value. I tried to do that first, by like storing a copy of the old value before the ban and then restoring it later, but the more I played around with that design the more hacky it became and I would run into concerns about retention

The way I've done it now removes that issue entirely since the new APIs don't actually touch the data of the old APIs, and everything just works seamlessly now

src/services/grpc/account/v2/update-server.ts
```diff
@@ -1,9 +1,223 @@
...
+			server.device = device;
+		}
+
+		if (aesKey !== undefined) {
+			if (aesKey.length === 0) {
+				throw new ServerError(Status.INVALID_ARGUMENT, 'AES key cannot be empty');
+			}
+
+			server.aes_key = aesKey;
+		}
```
DaniElectra: The GRPC token system doesn't require AES keys, so this doesn't apply anymore

jonbarrow: I thought we were still reusing the AES key for things? I remember reusing the AES key for something in the friends server but can't remember off the top of my head what

DaniElectra: The AES keys on friends are for the pidHMAC calculation and the Mii decryption. Other games that were ported to the latest lib versions don't have any keys

jonbarrow: Oh right, got ya

